PR #1, opened 2018-11-13
legacy-support: change comment style

kencu:
suppresses dozens of warning messages for C++ style comments
not being appropriate for older C standards

### DIFF
--- a/include/MacportsLegacySupport.h
+++ b/include/MacportsLegacySupport.h
@@ -20,27 +20,27 @@
 
 #include "AvailabilityMacros.h"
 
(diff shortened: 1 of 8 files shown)


cjones051073:
LGTM. 

Just out of interest though, what OS/compiler are you seeing warnings on ? I never saw any myself during testing.

jmroot:
Might have been easier to just build in C99 mode, but oh well.

kencu:
Indeed, I thought about that, adding that flag for the libuv tests...but this way, it's fixed in all ports forever, so seemed the most compatible way to get to "it just works"...

PR #2, opened 2018-11-21
add support for *at calls

kencu:
uses Jeremy's Apple Open Source implementation of userland *at functions
lots of different functions and definitions added to multiple headers
but it seems to all hang together and was successful at building `util-linux` in the current version on < 10.10, which previously did not build.


cjones051073:
On a quick look, this all looks fine to me... thanks !

kencu:
I just looked it all over again, and don't see any glaring errors either. It works for me here. I guess we'll let it out into the wild for people to try and then include it in the next release once it passes a few hurdles.

kencu:
Thanks, @jeremyhu !

cjones051073:
To make it available a release will need to be made. So far I have been calling them candidate pre-releases, 

https://github.com/macports/macports-legacy-support/releases

and numbering them 0.x, but perhaps its time to start at 1.0, now this package is under the MP organisation ?

PR #4, opened 2018-12-14
time: add 'clockid_t'

michaelld:
`clockid_t` is the actual type of the first argument to both `clock_*` functions; it might be used by the calling code; just typedef it to be an `int` for easy compatibility. This change works for me to fix building `bladeRF` on OSX 10.5 PPC, which definitely does -not- have the `clock_*` functions.


cjones051073:
LGTM. @kencu ?

kencu:
yep, looks fine. thanks!

PR #5, opened 2018-12-14
Add <wchar.h> function wcpcpy

ccorn:
The `<wchar.h>` function [`wcpcpy`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/wcpcpy.html) is needed e.g. for building port `mpd`.
It is definitely missing in darwin 9 (Mac OS X 10.5) / PPC.
I [suspect](http://www.manpagez.com/man/3/wmemmove/) that it is present in darwin 12 (Mac OS X 10.8) although it seems not to be covered in Apple's [manpage archive](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/wcscpy.3.html).
I am assuming introduction of `wcpcpy` in darwin 11 (Mac OS X 10.7). Please verify.

Implementation notes:
I looked at the [GLIBC implementation](https://code.woboq.org/userspace/glibc/wcsmbs/wcpcpy.c.html) and did not like its taste, neither that of the [simple version used for benchmarking](https://code.woboq.org/userspace/glibc/benchtests/bench-stpcpy.c.html), so I created my own version instead. String operations are memory-bound, so the differences in the pointer arithmetic steps should not matter much. Using `__restrict` allows compiler optimizations like batch prefetching without the need to check for overlaps.


kencu:
Hey! Thanks for the help! In fact I was working on just this this morning, and I did a branch with all of them. Please check this out <https://github.com/macports/macports-legacy-support/tree/ws> and see if that meets your needs. I tried to find every one we are missing and added it in.

ccorn:
> Please check this out https://github.com/macports/macports-legacy-support/tree/ws and see if that meets your needs. I tried to find every one we are missing and added it in.

Good idea. Some refinements suggested in [a comment there](https://github.com/macports/macports-legacy-support/commit/302bbcc9c8593eb15b5e972d989f139405c8ccd0#commitcomment-31687622).

ccorn:
Closing this while preparing a more encompassing PR in the [`wcs` branch](https://github.com/ccorn/macports-legacy-support/tree/wcs).

PR #6, opened 2018-12-15
Add more `wc*` functions missing in OS X 10.6 and earlier.

ccorn:
This has evolved from a combination of #5 and kencu's [`ws` work](https://github.com/macports/macports-legacy-support/commit/302bbcc9c8593eb15b5e972d989f139405c8ccd0#commitcomment-31688247).

Some notes:
* Updated `README.md` to cover the newly provided functions.
* No need for implementations of `wcsxfrm_l` and `wcscoll_l`: Darwin 9 has them (in `<xlocale/_wchar.h>`). Remember that `<xlocale.h>` must be included *after* the headers that shall be made locale-aware; otherwise you won't get the associated `*_l` function declarations.
* Added a test file for the `wcs{,n}casecmp` implementations. Not really much, but better than nothing.
* The Makefile has been completely rewritten. It is now more straightforward and supports testing (`make test` or `make check`) in the build tree. Adding tests is easy: Simply create files matching `test/test_*.c` or `test/test_*.cpp`, then running `make check` will compile them, link them with the Legacy support library, and run them, possibly aborting the `make` run in case of non-zero exit status.
* Upon installation, the dylib gets postprocessed with `install_name_tool` to change its identifier to the final installation path.
* Added parentheses to `__MP_LEGACY_SUPPORT_*` macro defs for safer usage.
* Partitioned the set of the new `wc*` functions into subsets of closely related functions where each subset is controlled by a distinct macro. Thus we still have expressive macro names and quite fine-grained control.
* I have slapped an MIT-style license on source files that are new or completely rewritten (as with `Makefile`). In particular, `src/wcsnlen.c` contains code for `wcsnlen` which I suppose comes from @kencu (it's [in his `ws` branch](https://github.com/macports/macports-legacy-support/commit/302bbcc9c8593eb15b5e972d989f139405c8ccd0#diff-f8d02dd2910d2c1e09c0281e6061627cR54)), therefore Ken is specified as that file's copyright holder now. The others carry my name, even though I have tried to mimic the idioms used in nearby source files.
* I have *not* revised the already-provided `wcsdup` implementation (which, as you will easily notice, is susceptible to overflow).
* Does it work? Well, I can now build the port `mpd` with `PortGroup legacysupport` which would otherwise fail for lack of `wcpcpy`.


kencu:
Thank you! I'll get a few of us @cjones051073 @michaelld and me to look it over and then it looks like a great addition!

src/wcsnlen.c
```diff
@@ -0,0 +1,31 @@
+
+/*
+ * Copyright (c) 2018 Ken Cunningham
+ *
```
kencu: I co-opted this right out of musl, so if you want to replace it with yours go right ahead. Otherwise we should credit them (MIT license).

Makefile
```diff
@@ -1,6 +1,5 @@
-
-
-# Copyright (c) 2010 Chris Jones <jonesc@macports.org>
+# GNU Makefile for MacportsLegacySupport
+# Copyright (c) 2018 Christian Cornelssen
 #
```
kencu: For all the Makefile stuff I defer to @cjones051073  -- might be nice to leave him credited, though, even though you  rewrote it, given the inspiration and collegiality of it all. What do you think?

src/wcsnlen.c
ccorn: Ah, [I see](https://git.musl-libc.org/cgit/musl/tree/src/string/wcsnlen.c). I'll work in [their text](https://git.musl-libc.org/cgit/musl/tree/COPYRIGHT).

jmroot:
> Upon installation, the dylib gets postprocessed with install_name_tool to change its identifier to the final installation path.

Why not link with the final install_name in the first place? If it's because the tests won't work without installing first, setting DYLD_LIBRARY_PATH is the usual way to fix that.

kencu:
Christian -- keep your credit in too! 10 years from now we can look back and smile!!

ccorn:
> Why not link with the final install_name in the first place? If it's because the tests won't work without installing first, setting DYLD_LIBRARY_PATH is the usual way to fix that.

Simple. *The library, as built, is not installed yet.* It should not pretend otherwise. The first use case after building is testing, not installation. If testing requires kludges like `*PATH` tweaks (where everyone seems to think the variables are only there for him, and others will never find conflicting uses for them), development is hindered. And that's not even necessary: Setting the library up so that it is perfectly suitable for its current location is easy; one `install_name` adaptation does the job, and this is precisely what the build tools should do automatically. Now, `make install` creates a new instance of the library in another location, so it is perfectly reasonable to adapt the new instance accordingly. *Both* versions of the library (buildtree and system) are then usable, and that is how things should be.

jmroot:
Using install_name_tool is just another kludge, but whatever, as long as -headerpad_max_install_names is used it works.

kencu:
Well this looks good. Why don't we merge it into the branch, at least, and then we can work on it further and test it from there ?

cjones051073:
No problem. My question wrt versioning was more in general anyway, not really related to this.

PR #7, opened 2018-12-16
Makefile: tests separate compile and link steps, honoring LDFLAGS

ccorn:
The current state is that test programs are compiled and linked in one go, using `CFLAGS` resp. `CXXFLAGS` and the linker options for locating the buildtree library, but no LDFLAGS.
This is no problem if the only important thing in the LDFLAGS are `-arch` options because those are presumably duplicated in the `C*FLAGS` (at least in the Portfile).
Yet this seems inappropriate to me, so I have moved the `-L` part from the contents of `TESTLIBS` into a new variable `TESTLDFLAGS` which now also has `$(LDFLAGS)` appended.
With this simple change, the compile-and-link rule for test programs uses both `C*FLAGS` and `LDFLAGS`. This can lead to duplicated `-arch` options. While this should not be a problem in practice, I do not like the look of that.
Therefore I have now split the test build rules into separate compilation and linking steps.
As a result, the `test` subdirectory now receives additional build artifacts which need to be cleaned up by `make clean`. Therefore I have updated that rule as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,9 +51,12 @@ LIBOBJECTS      := $(patsubst %.c,%.o,$(wildcard $(SRCDIR)/*.c))
 TESTDIR          = test
 TESTNAMEPREFIX   = $(TESTDIR)/test_
 TESTRUNPREFIX    = run_
-TESTLIBS         = -L$(BUILDLIBDIR) -l$(LIBNAME)
+TESTLDFLAGS      = -L$(BUILDLIBDIR) $(LDFLAGS)
+TESTLIBS         = -l$(LIBNAME)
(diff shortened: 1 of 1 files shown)


cjones051073:
Seems fine to me. Do you agree @kencu ?

As an aside, I am thinking it is perhaps time for a new 0.7 release ?

michaelld:
I like the idea behind this change. Haven't tested it to verify functionality. Once this is verified & merged, I'd highly recommend a new release.

kencu:
Looks fine -- my Makefile fu is only basic, you three are ahead of me there. Thanks for all the input from everyone, this is making my life much easier.

cjones051073:
I'll cut a new 0.7 release later on this evening, unless someone beats me to it...

PR #8, opened 2019-01-22
add "wrap" functions

kencu:
these changes "wrap" existing library functions
modifying the behaviour to behave like current
expected library functions

realpath: leopard and earlier
support NULL as buffer

sysconf: tiger
support _SC_NPROCESSORS_CONF and _SC_NPROCESSORS_ONLN


kencu:
@ccorn -- if you have a moment, could you look this over?

ccorn:
Some refinements for sysconf on [my version](https://github.com/macports/macports-legacy-support/compare/6fe7cf0...ccorn:wrapfunctions) of the `wrapfunctions` branch:

* On 10.5 (which is the first non-wrapped version), the extra sysconf parameter defines are conditioned on `!defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)` which makes sense (defining `_POSIX_C_SOURCE` should disable nonstandard features), so I added that condition to the definition of `__MP_LEGACY_SUPPORT_SYSCONF_WRAP__`.
* Define `_SC_NPROCESSORS_CONF`, `_SC_NPROCESSORS_ONLN` independently from `__BUILDING_MP_LEGACY_SUPPORT_SYSCONF_WRAP__`; this avoids the need for redefinition in `macports_legacy_sysconf.c`.
* Some safety measures (checking return value of `sysctl` and presetting its presumed output variable)

cjones051073:
Looks OK to me. As always, your really best suited to test anything that only affects really old systems, like 10.4. As long as on newer systems the changes have no impact, and it looks like that is all setup correctly, the the impact of any issues related to this is rather isolated...

ccorn:
Any suggestions for software that needs the new wrappers?

ccorn:
Added a minimal testcase for `realpath`.

ccorn:
Updated `README.md`. If you like those changes, you can pull [ccorn:wrapfunctions](https://github.com/ccorn/macports-legacy-support/tree/wrapfunctions) into this pull request branch.

ccorn:
Moved some `#endif` so that the `__BUILDING_*` macros only guard the redefinitions, not the declarations of the macports wrappers. Thus the compiler can check for discrepancies between the declarations and later definitions.

ccorn:
Added minimal test for the sysconf wrapper. Not tested on 10.4 where it is supposed to take effect. Please `git pull https://github.com/ccorn/macports-legacy-support wrapfunctions` and `make run_sysconf` on 10.4 (and `make run_realpath` while you are at it).
Should be good to go then.

kencu:
> Any suggestions for software that needs the new wrappers?

yes, right off, gegl-0.4 (and others) crash calling realpath with a NULL buffer. Fixing that allows gimp2 to finish on PPC with a few tweaks.

Thanks to both of you for your time and knowledge. I'll look over and test your suugestion, Christian. I may alter some a bit as for these systems my prime goal is maximal compatibliity and not so much POSIX purity. 

I have a few more coming once we settle on the idea that this is the proper method to accomplish this function wrapping. Looks like tha's a go.

ccorn:
Note that if you remove the _POSIX_C_SOURCE etc, then compiling software with that macro would fail under 10.5 and later (as it should) but not under 10.4.

ccorn:
On the other hand, macports-legacy-support is *supposed* to be applied to sources which need it, and those would not define _POSIX_C_SOURCE. So in practice that sublety should not matter.

ccorn:
Of you remove the _POSIX_C_SOURCE stuff, please do me a favor and remove the (ugly) counteracting `#undef _POSIX_C_SOURCE` in `src/macports_legacy_sysconf.c` as well.

kencu:
Last time I looked through the `10.5` headers, almost overwhere where they blocked part of it the a `#ifndef _POSIX_C_SOURCE` blocker, they OR with a `_DARWIN_C_SOURCE` enabler, which is almost always defined. So -- pointless, really, on Leopard. But this causes huge issues on Tiger, where the `_DARWIN_C_SOURCE` enabler doesn't exist in the headers, so lots of stuff breaks.

kencu:
I read about the `__restrict` keyword as a compiler assistance device -- interesting indeed. But it seems to be only available in `c99` mode or newer based on my Googling? Do they #define it out somehow in the 10.4 and 10.5 headers if the c standard is older than `1999`?

ccorn:
Yes, look at `sys/cdefs.h`.

ccorn:
Also note that `restrict` is a qualifier in the same category as `const` and `volatile` which implies that function signatures with or without that qualifier applied to some arguments are compatible. Therefore, the following compiles:

~~~ c
typedef char *(*A)(const char *);
typedef char *(*B)(const char * restrict);
typedef char *(*C)(const char * const);

A ab(B b) {return b;}
B ba(A a) {return a;}
A ac(C c) {return c;}
C ca(A a) {return a;}
B bc(C c) {return c;}
C cb(B b) {return b;}
~~~
This means that we can safely use `__restrict` even on 10.4 even though 10.4 `realpath` does not use it whereas 10.5+ `realpath` does.

rmottola:
Just to be pedantic... I would use 
*tmp = calloc(PATH_MAX, sizeof(char*));
instead of just malloc(PATH_MAX). With char* sizeof it shouldn't be a matter...

kencu:
OK. I'm going to commit this then, and pull in the fixes Christian did after. Thanks!

PR #9, opened 2019-01-25
Wrapfunctions2

ccorn:
Regarding the wrapping pattern, I recommend to use the `#ifndef __BUILDING_*` guards only for preventing the macro (re)definitions, nothing else. Firstly because that is their actual purpose, secondly because otherwise those guards filter out too much, e.g. the definitions of the new `_SC_*` constants (so you have to duplicate them in the `.c` file) and the wrapper function prototype declaration (which prevents the compiler from checking for inconsistencies).
This pull request implements that proposal and fixes some whitespace issues along the way (notably, some tabs in `test/test_sysconf.c` are converted to 4 spaces each.)


ccorn:
Wrong branch point?

ccorn:
Repaired (I had a diverging master branch).

cjones051073:
Thanks. Please though can you clean up, squash, the commits. You currently have a merge commit in there, which I think we should avoid ending up in the project history.

PR #10, opened 2019-01-25
Use __BUILDING_* macros to prevent macro (re)definitions only.

ccorn:
(Repaired version of #9 which cannot be made to look clean by force-pushing)

Regarding the wrapping pattern, I recommend to use the `#ifndef __BUILDING_*` guards only for preventing the macro (re)definitions, nothing else. Firstly because that is their actual purpose, secondly because otherwise that pattern filters out too much, e.g. the definitions of the new `_SC_*` constants (so you have to duplicate them in the `.c` file) and the wrapper function prototype declaration (which prevents the compiler from checking for inconsistencies).

I have also fixed some tab/whitespace issues in `test/test_sysconf.c` along the way.

### DIFF
--- a/include/stdlib.h
+++ b/include/stdlib.h
@@ -26,13 +26,13 @@
 /* realpath wrap */
 #if __MP_LEGACY_SUPPORT_REALPATH_WRAP__
 /* we need a way to isolate ourselves from our header wrapping while building */
-#ifndef __BUILDING_MP_LEGACY_SUPPORT_REALPATH_WRAP__
 
+#ifndef __BUILDING_MP_LEGACY_SUPPORT_REALPATH_WRAP__
 /* we are going to move the old realpath definition out of the way */
 #undef realpath
 #define realpath realpath_macports_original
-
 #endif /*__BUILDING_MP_LEGACY_SUPPORT_REALPATH_WRAP__ */
+
 #endif /*__MP_LEGACY_SUPPORT_REALPATH_WRAP__*/
 
 
@@ -42,11 +42,12 @@
 
 /* realpath wrap */
 #if __MP_LEGACY_SUPPORT_REALPATH_WRAP__
-#ifndef __BUILDING_MP_LEGACY_SUPPORT_REALPATH_WRAP__
 
+#ifndef __BUILDING_MP_LEGACY_SUPPORT_REALPATH_WRAP__
 /* and now define realpath as our new wrapped function */
 #undef realpath
 #define realpath macports_legacy_realpath
+#endif /*__BUILDING_MP_LEGACY_SUPPORT_REALPATH_WRAP__ */
 
 #ifdef __cplusplus
 extern "C" {
@@ -55,7 +56,6 @@ extern "C" {
 #ifdef __cplusplus
 }
 #endif
-#endif /*__BUILDING_MP_LEGACY_SUPPORT_REALPATH_WRAP__ */
 #endif /*__MP_LEGACY_SUPPORT_REALPATH_WRAP__*/
 
 
--- a/include/unistd.h
+++ b/include/unistd.h
@@ -3,13 +3,13 @@
 
 
 #if __MP_LEGACY_SUPPORT_SYSCONF_WRAP__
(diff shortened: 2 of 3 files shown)


mojca:
Could you please split the whitespace changes from the rest (i.e. make two commits)?

ccorn:
I have removed the whitespace changes for now. I can add those in a new commit here (although that is unrelated), but an independent branch+PR for pure whitespace changes seems more appropriate.

ccorn:
The whitespace changes are [in my ws branch](https://github.com/ccorn/macports-legacy-support/tree/ws) now.

kencu:
thanks. I'll take a look again, but actually I pulled in the parts I wanted and I might let the rest be for now, and see how this works out.

The essence is the same. I just understand nicely how I have it now, and I want to try to keep it separate in my mind. I build these to be built and tested sepate of the headers. 

If I don't take in everything you had suggested, please don't be discouraged. The big strokes are what I'm after, and your looking it over is / was much appreciated!

Next: To try getting libdispatch working on <10.6!!

Best, Ken

kencu:
This function wrapping stuff can be a bit mind- bending. I find I often have to pause for a moment to make sure what frame of reference I'm in. Am I wrapped, or standard? What about other defs? What is currently defined as what?

So to make sure I don't make errors, I develop these standalone. Keeping it separate of the header defs helps me keep it straight. I see where you're coming from with the header defs thing; perhaps when I've done this longer it will seem more natural to me to do it that way.

kencu:
I see your point, but I think for now we can leave it like this while I get used to it all.

Another thing on my mind, that only comes from being around MacPorts for a couple of years -- in a month, or three, someone is going to have an issue building some port or another on one of these systems. And when that happens, there will be some blowback about this "function wrapping" mechanism. Possibly substantial blowback.

I would like a "no doubt" #define I can use / give to someone else to use that disables EVERYTHING in the wrapping mechanism. That's also why I put it like this, in that location. In fact, I might add another #define test, something like "DISABLE_ALL_MACPORTS_FUNCTION_WRAPPING" that applies to both of these wraps and any other wraps we come up with, exactly for this blowback reason. I'll put all three tests right at the start of the wrapping blocks, so there can be no doubt about any of it.

I'll fix up the whitespace issue separately.

By the way, can you help me have the Makefile also make a second parallel static library of libMacPortsLegacySupport, so people who want these functions in statically can do it that way if they prefer?

michaelld:
it would be a good idea to be able to disable -all- wrapping, as well as being able to disable each wrapped function independently. I can see a situation where some wrapping works but others doesn't, so you can't disable all wrapping, but want to disable some wrapping.

michaelld:
So I take it that a #define has higher precedent than an actual function declaration, and that's how the wrapping works? Hmmm ... I guess that makes sense because the macro pre-processor comes before the compiler. Hmmmm ... that's very clever!

kencu:
I have no illusions that it always works, though. There must be some situations where this fails -- there always are! Time will tell. Just trying to make life as easy as possible, so ideally people don't even need to think about "older systems" -- they Just Work.

michaelld:
+1 and "amen"!

ccorn:
> By the way, can you help me have the Makefile also make a second parallel static library of libMacPortsLegacySupport, so people who want these functions in statically can do it that way if they prefer?

Take a look at [ccorn:slib](https://github.com/macports/macports-legacy-support/compare/b7d3b040d286f096d4ca75e589f352b70273d930...ccorn:slib) and its parent branch [ccorn:build](https://github.com/ccorn/macports-legacy-support/tree/build):

    git fetch https://github.com/ccorn/macports-legacy-support slib:slib

kencu:
@ccorn -- I pulled in your Makefile fixes, and thank you very much! I decided for now to go a slightly different, hopefully in the end more clear way with the headers. I agree, it requires I redundantly declare those two needed defines in the source. For the moment, I will live with that in the hope for future clarity! Thanks for all your contributions! Hopefully the static library will prove useful for some.  I will close this PR then, given the change in direction for now. -- Ken

PR #12, opened 2019-02-01
fix various compilation warnings

cjones051073:
tested on OSX10.6 and macOS 10.14. No warnings on either.

### DIFF
--- a/include/MacportsLegacySupport.h
+++ b/include/MacportsLegacySupport.h
@@ -60,18 +60,26 @@
 #define __MP_LEGACY_SUPPORT_POSIX_MEMALIGN__  (__APPLE__ && __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1060)
 
 /*  realpath() wrap */
-#define __MP_LEGACY_SUPPORT_REALPATH_WRAP__   (__APPLE__ && __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1060      \
-                                                         && !defined(__DISABLE_MP_LEGACY_SUPPORT_REALPATH_WRAP__)    \
-                                                         && !defined(__DISABLE_ALL_MACPORTS_FUNCTION_WRAPPING__) )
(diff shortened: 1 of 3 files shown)


kencu:
Thanks -- I'll check it on gcc-4.2, on Tiger -- it's back there that things always seem to show up :>

kencu:
I was trying that kind of thing last night -- there is something about the way this works on Tiger makes the realpath_test.c fail (the name doesn't get substituted) whereas with the previous one it works OK. It doesn't get to the sysconf  test so I'm not sure about that one.

I'll play with it more today.

kencu:
Same on Leopard. With the current header tests, everything works, but we're getting Wexpansion-to-defined warnings on clang that I need to get rid of.

With this version of the headers, the wrapping doesn't seem to take hold in the tests -- and at this moment I'm not sure why that might be. It looks like it should work, but it doesn't...

kencu:
The problem is not with the test, but with my blocker #define in the macports_legacy_realpath.c source. With this version of the headers, the blocker is not taking hold, and results in this mess (infinite recursion). 
```
char *
macports_legacy_realpath(const char * stringsearch, char * buffer)
{
    if (buffer == ((void *)0)) {
        char *myrealpathbuf = malloc(1024);
        if (myrealpathbuf != ((void *)0)) {
            return(macports_legacy_realpath(stringsearch,myrealpathbuf));
        } else {
            return(((void *)0));
        }
    } else {
        return(macports_legacy_realpath(stringsearch,buffer));
    }
}
```
Now -- why does the other version of the include guards work, but this version not?...

cjones051073:
LGTM. still no warnings on OSX10.6 or macOS10.14.

cjones051073:
p.s. please feel free to merge, once you are happy on < 10.6

kencu:
Yep -- works here on 10.4, 10.5, and above. I still would like to better understand why this #undef is needed now but not before, but hey, life is short, and time to move on! No warnings, everybody happy. Let's proceed.

PR #13, opened 2019-02-05
Fix c++11 <cmath> PPC 10.[45], GNU g++ 4.6 through 6 only

michaelld:
Fix c++11 <cmath> PPC 10.[45], GNU g++ 4.6 through 6 only, by adding the following line of code -before- any of the GNU header are included:
`#define _GLIBCXX_USE_C99_MATH_TR1`

Also: <math.h> is for C declarations, while <cmath> is for C++ declarations, so put the correct declarations in the correct header. <cmath> includes <math.h>, so this change should not effect current ports that rely on this legacy header.

Add modeline to <cmath>; probably some whitespace fixes too.

### DIFF
--- a/include/math.h
+++ b/include/math.h
@@ -1,0 +1,104 @@
+
+/*
+ * Copyright (c) 2010 Chris Jones <jonesc@macports.org>
+ * Copyright (c) 2019 Michael Dickens <michaelld@macports.org>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#ifndef _MACPORTS_MATH_H_
+#define _MACPORTS_MATH_H_
+
+/* MP support header */
+#include "MacportsLegacySupport.h"
+
+#if __MP_LEGACY_SUPPORT_LLROUND__
+
+/* For definition of __*_DECLS */
+#include <sys/cdefs.h>
+
+__BEGIN_DECLS
+
+/*
+ * These functions are present in the system math library but their
+ * prototypes might not be declared under some circumstances. Declare
+ * them here anyway.
+ */
+
+/*
+ * this is the same condition that defines the function prototypes in
+ * the system <math.h>.
+ */
+#if ( defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L ) || ! defined( __STRICT_ANSI__ )  || ! defined( __GNUC__ )
+#else
+extern long long int llrint   ( double );
+extern long long int llrintf  ( float );
+extern long long int llrintl  ( long double );
+
+extern long long int llround  ( double );
+extern long long int llroundf ( float );
+extern long long int llroundl ( long double );
+#endif
+
+__END_DECLS
+
+#endif /* __MP_LEGACY_SUPPORT_LLROUND__ */
+
+/*
+ * Include the next math.h, which might be from the primary system or
+ * it might be within GCC's c or c++ (yup!) headers
+ */
+
+/*
+ * If the GCC <math.h> header exists, then tell it: (1) to include the
+ * next <math.h>, which should be from the system; and (2) to not use
+ * it's <math.h> yet, because it basically wraps <cmath> and we need
+ * to keep everything herein focused on just <math.h>. If the user
+ * wants <cmath>, they should #include that specific header.
+ */
+
+/* store prior values for _GLIBCXX_MATH_H and _GLIBCXX_INCLUDE_NEXT_C_HEADERS */
+
+#ifdef L_GLIBCXX_MATH_H
+#undef L_GLIBCXX_MATH_H
+#endif
+#ifdef _GLIBCXX_MATH_H
+#define L_GLIBCXX_MATH_H _GLIBCXX_MATH_H
+#endif
+#define _GLIBCXX_MATH_H 1
+
+#ifdef L_GLIBCXX_INCLUDE_NEXT_C_HEADERS
+#undef L_GLIBCXX_INCLUDE_NEXT_C_HEADERS
+#endif
+#ifdef _GLIBCXX_INCLUDE_NEXT_C_HEADERS
+#define L_GLIBCXX_INCLUDE_NEXT_C_HEADERS _GLIBCXX_INCLUDE_NEXT_C_HEADERS
+#endif
+#define _GLIBCXX_INCLUDE_NEXT_C_HEADERS 1
+
+#include_next <math.h>
+
+/* restore prior values for _GLIBCXX_MATH_H and _GLIBCXX_INCLUDE_NEXT_C_HEADERS */
+
+#undef _GLIBCXX_MATH_H
+#ifdef L_GLIBCXX_MATH_H
+#define _GLIBCXX_MATH_H L_GLIBCXX_MATH_H
+#undef L_GLIBCXX_MATH_H
+#endif
+
+#undef _GLIBCXX_INCLUDE_NEXT_C_HEADERS
+#ifdef L_GLIBCXX_INCLUDE_NEXT_C_HEADERS
+#define _GLIBCXX_INCLUDE_NEXT_C_HEADERS L_GLIBCXX_INCLUDE_NEXT_C_HEADERS
+#undef L_GLIBCXX_INCLUDE_NEXT_C_HEADERS
+#endif
+
+#endif /* _MACPORTS_MATH_H_ */
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.o
 *.d
 *.dylib
+lib
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 # GNU Makefile for MacportsLegacySupport
 # Copyright (c) 2010 Chris Jones <jonesc@macports.org>
+# Copyright (c) 2019 Michael Dickens <michaelld@macports.org>
 #
 # Permission to use, copy, modify, and distribute this software for any
 # purpose with or without fee is hereby granted, provided that the above
@@ -111,6 +112,28 @@ $(TESTPRGS_C): %: %.o $(BUILDDLIBPATH)
 $(TESTPRGS_CPP): %: %.o $(BUILDDLIBPATH)
 	$(CXX) $(TESTLDFLAGS) $< $(TESTLIBS) -o $@
 
+# Special clause for testing the cmath fix: Just need to verify that
+# building succeeds or fails, not that the executable runs or what it
+# produces.  Note that for some reason all Clang compilers tested
+# (Apple and MP) successfully compile and link this code regardless of
+# the c++ standard chosen, which seems to be a build issue since the
+# functions being tested were not introduced until c++11. GCC
+# correctly fails the compile and link using c++03 or older, but
+# succeeds using c++11 -- as desired.
+test_cmath: test/test_cmath.cc $(ALLHEADERS)
+	$(info 1: testing compiler '$(CXX)' for non-legacy cmath using c++03; the build should fail, regardless of the compiler or OS)
+	$(info 1: $(CXX) $(CXXFLAGS) -std=c++03 $< -o test/$@_cxx03)
+	@-$(CXX) -I$(SRCINCDIR) $(CXXFLAGS) -std=c++03 $< -o test/$@_cxx03 &> /dev/null && echo "1: c++03 no legacy cmath build success (test failed)!" || echo "1: c++03 no legacy cmath build failure (test succeeded)!"
+	$(info 2: testing compiler '$(CXX)' for non-legacy cmath using c++03; the build should fail, regardless of the compiler or OS)
+	$(info 2: $(CXX) -I$(SRCINCDIR) $(CXXFLAGS) -std=c++03 $< -o test/$@_cxx03)
+	@-$(CXX) -I$(SRCINCDIR) $(CXXFLAGS) -std=c++03 $< -o test/$@_cxx03 &> /dev/null && echo "2: c++03 legacy cmath build success (test failed)!" || echo "2: c++03 legacy cmath build failure (test succeeded)!"
+	$(info 3: testing compiler '$(CXX)' for non-legacy cmath using c++11; if the compiler supports this standard, then the build should succeed regardless of OS)
+	$(info 3: $(CXX) $(CXXFLAGS) -std=c++11 $< -o test/$@_cxx11)
+	@-$(CXX) $(CXXFLAGS) -std=c++11 $< -o test/$@_cxx11 &> /dev/null && echo "3: c++11 no legacy cmath build success (test failed)!" || echo "3: c++11 no legacy cmath build failure (test succeeded)!"
+	$(info 4: testing compiler '$(CXX)' for legacy cmath using c++11; if the compiler supports this standard, then the build should succeed regardless of OS)
+	$(info 4: $(CXX) -I$(SRCINCDIR) $(CXXFLAGS) -std=c++11 $< -o test/$@_cxx11)
+	@-$(CXX) -I$(SRCINCDIR) $(CXXFLAGS) -std=c++11 $< -o test/$@_cxx11 &> /dev/null && echo "4: c++11 legacy cmath build success (test succeeded)!" || echo "4: c++11 legacy cmath build failure (test failed)!"
+
 $(TESTRUNS): $(TESTRUNPREFIX)%: $(TESTNAMEPREFIX)%
 	$<
 
@@ -134,12 +157,12 @@ install-slib: $(BUILDSLIBPATH)
 	$(MKINSTALLDIRS) $(DESTDIR)$(LIBDIR)
 	$(INSTALL_DATA) $(BUILDSLIBPATH) $(DESTDIR)$(LIBDIR)
 
-test check: $(TESTRUNS)
+test check: $(TESTRUNS) test_cmath
 
 clean:
 	$(RM) $(foreach D,$(SRCDIR) $(TESTDIR),$D/*.o $D/*.d)
-	$(RM) $(BUILDDLIBPATH) $(BUILDSLIBPATH) $(TESTPRGS)
+	$(RM) $(BUILDDLIBPATH) $(BUILDSLIBPATH) $(TESTPRGS) test/test_cmath_*
 	@$(RMDIR) $(BUILDDLIBDIR) $(BUILDSLIBDIR)
 
-.PHONY: all dlib slib clean check test $(TESTRUNS)
+.PHONY: all dlib slib clean check test $(TESTRUNS) test_cmath
 .PHONY: install install-headers install-lib install-dlib install-slib
--- a/README.md
+++ b/README.md
@@ -14,7 +14,8 @@ via the legacysupport PortGroup.
 
 Wrapped headers are:
 
- - cmath         : Adds declaration of various `long long` methods missing in OSX10.6 and older.
+ - cmath         : Adds the same functions as those provided by the herein math.h, in namespace std:: .
+ - math.h         : Adds declaration of various `long long` methods  missing in OSX10.6 and older, GCC 8 and older.
  - pthread.h     : Adds PTHREAD_RWLOCK_INITIALIZER for OSX10.4
  - stdio.h       : Adds `getline` and `getdelim` functions missing in OSX10.6 and older.
  - stdlib.h      : Adds `posix_memalign` functional replacement,
--- a/include/MacportsLegacySupport.h
+++ b/include/MacportsLegacySupport.h
@@ -2,6 +2,7 @@
 /*
  * Copyright (c) 2010 Chris Jones <jonesc@macports.org>
  * Copyright (c) 2018 Ken Cunningham <kencu@macports.org>
+ * Copyright (c) 2019 Michael Dickens <michaelld@macports.org>
  *
  * Permission to use, copy, modify, and distribute this software for any
  * purpose with or without fee is hereby granted, provided that the above
@@ -81,5 +82,13 @@
 /* pthread_rwlock_initializer */
 #define __MP_LEGACY_SUPPORT_PTHREAD_RWLOCK__  (__APPLE__ && __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1050)
 
+/* c++11 <cmath> PPC 10.[45] and Intel 10.[4-6], GNU g++ 4.6 through 8. */
+#if (__APPLE__ && __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1070 \
+               && defined(__GNUC__) && (__GNUC__ <= 8)                 \
+               && ((__GNUC__ > 4) || ((__GNUC__ == 4) && (__GNUC_MINOR__ >= 6))))
+#define __MP_LEGACY_SUPPORT_CXX11_CMATH__ 1
+#else
+#define __MP_LEGACY_SUPPORT_CXX11_CMATH__ 0
+#endif
 
 #endif /* _MACPORTS_LEGACYSUPPORTDEFS_H_ */
--- a/include/cmath
+++ b/include/cmath
@@ -1,6 +1,6 @@
-
+/* -*- c++ -*- */
 /*
- * Copyright (c) 2010 Chris Jones <jonesc@macports.org>
+ * Copyright (c) 2019 Michael Dickens <michaelld@macports.org>
  *
  * Permission to use, copy, modify, and distribute this software for any
  * purpose with or without fee is hereby granted, provided that the above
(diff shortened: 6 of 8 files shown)


kencu:
Let me test some Intel systems...we had issues with gcc48 math on 10.5. And I previously had these math issues on 10.6 building webkit2-gtk against macports-libstc++ but that was when libgcc was provided by gcc6, so things might be different now....

cjones051073:
I guess LGTM, although I don't really have the machines needed to test this, so I will leave that to @kencu 

One small thought request. It would be nice if you could also add a little test application to test the math changes here. Basically nothing more complex that checking if the various `ll<xyz>` methods are found or not. It makes testing much easier if the inbuilt `make test` can be used, rather than having to test against some port or other. ( We should have been more careful about always adding test cases for each new feature from the start, but we didn't really think of it back then. )

michaelld:
@cjones051073 +1 & sure I can add a test ... very good idea!

FYI: I'm building cmake-devel right now using these changes ... fingers crossed it works! I'll also test out on other ports that use this PG to verify that the changes do no harm ... hopefully @kencu can do the same.

michaelld:
cmake-devel succeeded! Now on to other testing & getting a test in place.

kencu:
I'll check 10.4 to 10.6 Intel later today. Pretty sure they also need these fixes from past experience. 

I admit I can get a bit confused about which headers gcc is actually using sometimes -- e.g. when gcc5 is using libgcc8 on 10.5 Intel, exactly which folder of headers are actually being used???? libgcc8's installed headers, or gcc5's installed headers?

cjones051073:
> I'll check 10.4 to 10.6 Intel later today. Pretty sure they also need these fixes from past experience.
> 
> I admit I can get a bit confused about which headers gcc is actually using sometimes -- e.g. when gcc5 is using libgcc8 on 10.5 Intel, exactly which folder of headers are actually being used???? libgcc8's installed headers, or gcc5's installed headers?

Now that's easy to answer ;)

```
Titan ~/Projects/MacPorts/legacy-support > gcc-mp-8 -xc++ -E -v -
Using built-in specs.
COLLECT_GCC=gcc-mp-8
Target: x86_64-apple-darwin17
Configured with: /opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_lang_gcc8/gcc8/work/gcc-8.2.0/configure --prefix=/opt/local --build=x86_64-apple-darwin17 --enable-languages=c,c++,objc,obj-c++,lto,fortran --libdir=/opt/local/lib/gcc8 --includedir=/opt/local/include/gcc8 --infodir=/opt/local/share/info --mandir=/opt/local/share/man --datarootdir=/opt/local/share/gcc-8 --with-local-prefix=/opt/local --with-system-zlib --disable-nls --program-suffix=-mp-8 --with-gxx-include-dir=/opt/local/include/gcc8/c++/ --with-gmp=/opt/local --with-mpfr=/opt/local --with-mpc=/opt/local --with-isl=/opt/local --enable-stage1-checking --disable-multilib --enable-lto --enable-libstdcxx-time --with-build-config=bootstrap-debug --with-as=/opt/local/bin/as --with-ld=/opt/local/bin/ld --with-ar=/opt/local/bin/ar --with-bugurl=https://trac.macports.org/newticket --disable-tls --with-pkgversion='MacPorts gcc8 8.2.0_3'
Thread model: posix
gcc version 8.2.0 (MacPorts gcc8 8.2.0_3) 
COLLECT_GCC_OPTIONS='-E' '-v' '-mmacosx-version-min=10.13.0' '-asm_macosx_version_min=10.13' '-mtune=core2'
 /opt/local/libexec/gcc/x86_64-apple-darwin17/8.2.0/cc1plus -E -quiet -v -D__DYNAMIC__ - -fPIC -mmacosx-version-min=10.13.0 -mtune=core2
ignoring nonexistent directory "/opt/local/lib/gcc8/gcc/x86_64-apple-darwin17/8.2.0/../../../../../x86_64-apple-darwin17/include"
#include "..." search starts here:
#include <...> search starts here:
 /opt/local/include/gcc8/c++/
 /opt/local/include/gcc8/c++//x86_64-apple-darwin17
 /opt/local/include/gcc8/c++//backward
 /opt/local/lib/gcc8/gcc/x86_64-apple-darwin17/8.2.0/include
 /opt/local/include
 /opt/local/lib/gcc8/gcc/x86_64-apple-darwin17/8.2.0/include-fixed
 /usr/include
 /System/Library/Frameworks
 /Library/Frameworks
End of search list.
```

```
Titan ~/Projects/MacPorts/legacy-support > gcc-mp-6 -xc++ -E -v -
Using built-in specs.
COLLECT_GCC=gcc-mp-6
Target: x86_64-apple-darwin17
Configured with: /opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_lang_gcc6/gcc6/work/gcc-6.5.0/configure --prefix=/opt/local --build=x86_64-apple-darwin17 --enable-languages=c,c++,objc,obj-c++,lto,fortran --libdir=/opt/local/lib/gcc6 --includedir=/opt/local/include/gcc6 --infodir=/opt/local/share/info --mandir=/opt/local/share/man --datarootdir=/opt/local/share/gcc-6 --with-local-prefix=/opt/local --with-system-zlib --disable-nls --program-suffix=-mp-6 --with-gxx-include-dir=/opt/local/include/gcc6/c++/ --with-gmp=/opt/local --with-mpfr=/opt/local --with-mpc=/opt/local --with-isl=/opt/local --enable-stage1-checking --disable-multilib --enable-lto --enable-libstdcxx-time --with-build-config=bootstrap-debug --with-as=/opt/local/bin/as --with-ld=/opt/local/bin/ld --with-ar=/opt/local/bin/ar --with-bugurl=https://trac.macports.org/newticket --disable-tls --with-pkgversion='MacPorts gcc6 6.5.0_1'
Thread model: posix
gcc version 6.5.0 (MacPorts gcc6 6.5.0_1) 
COLLECT_GCC_OPTIONS='-E' '-v' '-mmacosx-version-min=10.13.7' '-asm_macosx_version_min=10.13' '-mtune=core2'
 /opt/local/libexec/gcc/x86_64-apple-darwin17/6.5.0/cc1plus -E -quiet -v -D__DYNAMIC__ - -fPIC -mmacosx-version-min=10.13.7 -mtune=core2
ignoring nonexistent directory "/opt/local/lib/gcc6/gcc/x86_64-apple-darwin17/6.5.0/../../../../../x86_64-apple-darwin17/include"
#include "..." search starts here:
#include <...> search starts here:
 /opt/local/include/gcc6/c++/
 /opt/local/include/gcc6/c++//x86_64-apple-darwin17
 /opt/local/include/gcc6/c++//backward
 /opt/local/lib/gcc6/gcc/x86_64-apple-darwin17/6.5.0/include
 /opt/local/include
 /opt/local/lib/gcc6/gcc/x86_64-apple-darwin17/6.5.0/include-fixed
 /usr/include
 /System/Library/Frameworks
 /Library/Frameworks
End of search list.
```

i.e. they use those specifically shipped for that compiler.

cjones051073:
which I guess then does beg the question, when are those shipped with `libgcc8` under `/opt/local/include/gcc/c++/` actually used...

michaelld:
OK I just pushed a new test & some gitignore tweaks.

The test has to have its own special clause, since it is different than the other tests.

kencu:
> when are those shipped with `libgcc8` under `/opt/local/include/gcc/c++/` actually used...

Right - that would be why previously I found that fixing the headers in libgcc8 would not fix this math issue for the older compilers.

kencu:
When the liquid-cooled processors in the G5's start leaking, that's when we bid them adieu.

kencu:
> when are those shipped with `libgcc8` under `/opt/local/include/gcc/c++/` actually used...

I think they are used by clang-3.9+ when building against -stdlib=macports-libstdc++.

kencu:
lots of math failures on 10.5 Intel with gcc48 and gcc6. Will tweak and retest.

kencu:
with gcc48 on 10.5 Intel, I see errors doing the test, but the final printout says "success".

michaelld:
OK well that's just strange on both accounts.

michaelld:
The oldest Intel OSX I have installed and working with MP is currently 10.6. I'll see if I see these issues with it & if so then I can probably figure out how to fix them. If not then I'll dig around & see about installing 10.5 Intel since that clearly is working for you to demonstrate this issue!

kencu:
FYI 10.5 Intel Server installs and runs nicely in Parallels. 10.4 Intel Server installs in VirtualBox with slight tweaking of vboxmanage.

I have hardware running 10.5 Intel (above). I'm not sure if I have an actual 10.4 Intel system set up anywhere -- don't think so.

michaelld:
I have an older MacBook laptop that I run 10.6 through 10.11 on & have a partition waiting for 10.5 ... just haven't gotten around to installing it. I do all of my testing on actual live hardware ... it's actually pretty straight forward to have a common login space & then different partitions, one for each OS version.

michaelld:
Thinking I have an OSX 10.5 install disk around somewhere. Hoping it still works :)

michaelld:
OK so on OSX 10.6, GCC6 fails the cmath test when using c++11 ... so, something for me to work with without having to spend hours on an OSX 10.5 install ... yay!?

michaelld:
@kencu Try with `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ <= 1068` instead ... that works for me & hopefully will for both of yours.

michaelld:
I'll look into the question about failed build but somehow "make test" shows success ... just strange!

michaelld:
OK so I can fix the "make test" printouts to work too. I'll push changes shortly & keep testing on my 10.6 install.

michaelld:
@kencu do you have GCC7 installed on 10.5 PPC? Can you test it out? I have 5.3.0_3 installed on 10.6 Intel and it needs the cmath fixes ... will update to 5.4.0_0 tonight & test that tomorrow, since maybe it has fixes in it already ... maybe? I'll get GCC7 building on PPC, but that'll take the weekend & so I can test early next week.

kencu:
Yea, all gccs <7 will need the full math fixes, and all gccs including 7+ need the llrint* fixes.

michaelld:
... but on only certain OSX versions. For example, on OSX 10.14, I've gone all the way back to GCC5 and it works correctly without needing the Legacy support fixes. I tend to believe that any OSX supporting libc++ will not need the fixes, whereas those using libstdc++ will require the fixes. Just a belief right now. Plenty of testing to be done!

kencu:
Yes, 10.7+ don't need any fixes as all the llrint* fcns are defined in math.h so gcc doesn't get confused.

michaelld:
OK good to know. So how do we determine inside our MacPortsLegacySupport.h header if the OS is the correct version? Hmmmm ....

kencu:
I'm still having some troubles with this for some reason...will have to start examining preprocessed source to understand why...

ccorn:
I will have to take a closer look. What I have noticed so far:

* The make `test_cmath` builds do not use any `LDFLAGS`. This is probably OK, as this is a header-only fix, so we do not need any special linker settings here, and archflags are typically given with `CXX` or `CXXFLAGS`.
* I suppose that you do not want that `__MP_LEGACY_SUPPORT_CXX11_CMATH__` includes the condition that `__cplusplus >= 201103L` (e.g. as a consequence of a `std=c++11` flag). Not doing this means that we get more from an older GCC than it typically enables, and our excuse is that the system library offers those functions anyway. Should we therefore look for other enhancement guards in the system headers and enable those as well? Where do we draw the line?

ccorn:
Discrepancy in `include/cmath`: `__MP_LEGACY_SUPPORT_CXX11_CMATH_PPC__` instead of `__MP_LEGACY_SUPPORT_CXX11_CMATH__`.

ccorn:
The `#define _GLIBCXX_USE_C99_MATH_TR1` in `include/cmath` should be guarded by an `#ifndef`, for obious reasons.

ccorn:
Turns out that `${prefix}/include/gcc/c++/cmath` tests for `_GLIBCXX_USE_C99_MATH_TR1` only if `__cplusplus >= 201103L`. If the usage is the same everywhere else, there is no feature creep, i. e. `-std=c++11` or later would still be required.

michaelld:
Thanks for looking into this issue further, @ccorn . I'll get back to this on Monday when I have access to my old build computers at work. Hopefully I'll be able to verify that the issue is only 10.6 and older, and, if so, then figure out a way to have these fixes work for those systems & cases only. I'll also fix the macro usage & protection as you note.

kencu:
This is proving to be a bit more of a PITA than I first thought it would be :>

We should remember that we still have Riccardo's backported gcc fixes that fix the root cause of this in gcc -- we could fairly easily just add those and revbump gcc < gcc7 to fix this as well... the only issue being a slight divergence from the stock gcc behaviour, but for the better, I would say.

kencu:
Sorry, michael, I messed you up slightly by merging the other PR.

michaelld:
No biggie. Pretty easy to fix the issue. I'll push an update here once I have it done.

michaelld:
I'm going to spend a little time this morning seeing if I can get this working correctly. I need it to get cmake-devel to build on PPC -- which means we'll need it for the next cmake release, too.

michaelld:
OK I've almost got a fix here. My discoveries / thoughts of the day thus far:

* `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__` is set for the specific GCC and OS, so it is usable to set a macro for just OSX 10.6 and earlier, and whatever version of GCC we wish.

* GCC7 and GCC8 internally check for the various `llround` and `llint` functions in the primary <math.h> ... but apparently do so using some flags that enable them in the main system <math.h> header. If you look at the eventual "math.h" header, you'll see that these functions are surrounded by an #if macro with various options.

* GCC7 and GCC8 write that these math functions are enabled into the c++config header that is read during compile time & thus in their provided <cmath> these functions are disabled. This is actually a bug in the way GCC handles these functions. GCC6 and prior do not do this check, so our initial work-around works; having to come up with something more creative for GCC7+ that works around their settings here.

I'm testing / verifying on 10.5 PPC, 10.6 Intel, 10.14, and probably one more 10.8+ Intel. Once I'm pretty confident, I'll force-push to this PR.

michaelld:
OK; changes pushed!

Some notes:

* the top-level system <math.h> does actually declare the "long long" functions under some circumstances for the -C- compiler. Since we're trying to fix them for the -c++- compiler, they will basically never be declared. That said, we can use the #if macro surrounding them correctly in our legacy math.h and cmath headers just in case.

* It looks like GCC9 fixes the c++ / c issue for when to declare the long long functions in std:: whereas GCC8 and GCC7 do not. GCC6 does include them as desired, but we still need the function prototypes from math.h. I haven't actually tested with GCC9 yet. I have tested on 10.5 PPC using GCC[5-7], 10.6 Intel using GCC[6-8], 10.9 using GCC[6-8], and 10.14 using GCC[6-8].

* I fixed the tests to display more relevant info when using clang.

* cmath is for c++ usage; math.h is for c usage.

ccorn:
After another cursory look, I suspect that intent and implementation of the first part of `include/cmath` differ:

~~~
#if __MP_LEGACY_SUPPORT_CXX11_CMATH__

#ifdef L_GLIBCXX_USE_C99_MATH_TR1
#undef L_GLIBCXX_USE_C99_MATH_TR1
#endif
#ifdef _GLIBCXX_USE_C99_MATH_TR1
#define L_GLIBCXX_USE_C99_MATH_TR1 _GLIBCXX_USE_C99_MATH_TR1
#endif

#define _GLIBCXX_USE_C99_MATH_TR1 1

#endif /* __MP_LEGACY_SUPPORT_CXX11_CMATH__ */

#include <math.h>
#include_next <cmath>

#undef _GLIBCXX_USE_C99_MATH_TR1
#ifdef L_GLIBCXX_USE_C99_MATH_TR1
#define _GLIBCXX_USE_C99_MATH_TR1 L_GLIBCXX_USE_C99_MATH_TR1
#undef L_GLIBCXX_USE_C99_MATH_TR1
#endif
~~~

* Nitpick: `#undef L_GLIBCXX_USE_C99_MATH_TR1` does not need to be guarded with an `#ifdef` of the same macro. It is specified to tolerate an undefined identifier.
* Even if `!__MP_LEGACY_SUPPORT_CXX11_CMATH__`, that `<cmath>` wrapper manipulates `_GLIBCXX_USE_C99_MATH_TR1`. Why? In that case, we do not want to manipulate things.
* Why explicit inclusion of `<math.h>` (and even if `!__MP_LEGACY_SUPPORT_CXX11_CMATH__`)? `#include_next <cmath>` includes it anyway (defining `_GLIBCXX_INCLUDE_NEXT_C_HEADERS` to prevent getting included from system's `<math.h>` for `__cplusplus`).
* If `__MP_LEGACY_SUPPORT_CXX11_CMATH__ && defined(_GLIBCXX_USE_C99_MATH_TR1)`, the wrapper ends up redefining `_GLIBCXX_USE_C99_MATH_TR1` to `L_GLIBCXX_USE_C99_MATH_TR1` literally.

Note that macro definitions do not expand their right hand side. The expansion occurs (recursively) at the point of invocation. You want to save and restore a macro definition? No way unless you use amazing elaborate tricks like [Boost.Preprocessor](https://www.boost.org/doc/libs/1_69_0/libs/preprocessor/doc/) does. But I do not suggest that we use BOOST here. Each of our patches should be kept small.

Suggestion: You can manually test your C++ preprocessing magic by doing e.g.

~~~
g++ -E -x c++ -Wall -Iinclude -D_GLIBCXX_USE_C99_MATH_TR1 - <<EOC | less
#include <cmath>
{_GLIBCXX_USE_C99_MATH_TR1}	// Just for looking at the token after expansion
EOC
~~~

michaelld:
Oh good catches there! I mean, the header will work as is, but it's not as desired for a wrapper header that may or may not do anything beyond wrapping. I'll fix your issues & test tomorrow & push those once verified.

michaelld:
I've tested / verified the updated <cmath> that hopefully meets all of our hopes and needs. It now passes through to the next cmath if legacy support is not required, without setting or modifying macros. I've verified building on 10.5 PPC / GCC and 10.6 Intel Clang / GCC, and it works as desired.

michaelld:
Related: In the test_cmath.cpp file, if using older Clang I have the code issue a warning that it might not compile ... which is true for `-stdlib=libstdc++` but not if `-stdlib=libc++`... I cannot find a way to figure out internal to the code which stdlib is being used! Any ideas from you folks?

kencu:
I found this, and it might be an appropriate test to see if `stdlib=libc++` in invoked:
```
# if !defined(_LIBCPP_VERSION)
```

kencu:
Yeah, well the opposite, obviously :> Missed the "!" on first paste.

ccorn:
Sorry for nitpicking again. There are a couple of slightly more substantial things to address, but I'll keep this post focused on the earlier topic.

Tl;dr: You *cannot* save and restore macro ``values'' as commented in `include/math.h` and apparently assumed in `include/cmath` as well. You *can* save and restore macro defined-ness. Your code happens to do that, despite the wrong underlying idea, but it still *looks* wrong. That can be easily repaired by simplifying some `#define` directives and removing mistaken comments.

Five pages (plus examples) from section 6.10.3 (Macro replacement) of the [C standard](http://www.open-std.org/JTC1/SC22/WG14/www/standards.html#9899) should serve to clarify my points. There have been almost no changes between C99 and C11 except for additional remarks. Essential point: Macro replacement happens not at the point of definition, but at the point of use, recursively. Looking at `include/cmath`:

~~~ c
#undef L_GLIBCXX_USE_C99_MATH_TR1
#ifdef _GLIBCXX_USE_C99_MATH_TR1
#define L_GLIBCXX_USE_C99_MATH_TR1 _GLIBCXX_USE_C99_MATH_TR1    // make this 1
#endif

#undef _GLIBCXX_USE_C99_MATH_TR1        // not in your code yet, but should be
#define _GLIBCXX_USE_C99_MATH_TR1 1
~~~
This does *not* store the previous ``value'' of `_GLIBCXX_USE_C99_MATH_TR1` in `L_GLIBCXX_USE_C99_MATH_TR1`. Indeed there is no general way to do such.

Depending on whether `_GLIBCXX_USE_C99_MATH_TR1` has been undefined or defined before, `L_GLIBCXX_USE_C99_MATH_TR1` will now be either undefined or expand recursively to whatever `_GLIBCXX_USE_C99_MATH_TR1` currently is (which is `1` right here, but changes later). That happens to suffice for our purposes, but the right-hand side in the definition of `L_GLIBCXX_USE_C99_MATH_TR1` should better be written as `1`. The advantage being that it will no longer track changes in `_GLIBCXX_USE_C99_MATH_TR1`. You never wanted subsequent changes in the latter to affect the former, right?

Later you do

~~~ c
#undef _GLIBCXX_USE_C99_MATH_TR1
~~~
Which means that now `L_GLIBCXX_USE_C99_MATH_TR1`, if defined as `_GLIBCXX_USE_C99_MATH_TR1`, would expand to exactly that latter token and not further.

~~~ c
#ifdef L_GLIBCXX_USE_C99_MATH_TR1
#define _GLIBCXX_USE_C99_MATH_TR1 L_GLIBCXX_USE_C99_MATH_TR1    // make this 1
#undef L_GLIBCXX_USE_C99_MATH_TR1
#endif
~~~
The condition is correct because the defined-ness of `L_GLIBCXX_USE_C99_MATH_TR1` equals the defined-ness of `_GLIBCXX_USE_C99_MATH_TR1` from before this `<cmath>` wrapper. The rest is broken so much that it happens to fit the typical use cases again:

If `_GLIBCXX_USE_C99_MATH_TR1` has been defined before including this `<cmath>` wrapper, then it now gets redefined to the token `L_GLIBCXX_USE_C99_MATH_TR1`. Upon use, it would not be expanded further because you `#undef` the latter. Thus the previous ``value'' of `_GLIBCXX_USE_C99_MATH_TR1` has been destroyed, but its defined-ness has been restored. Fortunately, only the defined-ness matters.

Should the `#undef L_GLIBCXX_USE_C99_MATH_TR1` be left out? Now that would be even funnier. Then `_GLIBCXX_USE_C99_MATH_TR1` would expand temporarily to `L_GLIBCXX_USE_C99_MATH_TR1` and then to `_GLIBCXX_USE_C99_MATH_TR1` again. At that point expansion would stop due to safeguards specified in the C standard, cf. 6.10.3.4(2). Thus, `_GLIBCXX_USE_C99_MATH_TR1` would expand to itself, but an `#ifdef` would still recognize it as defined. Which happens to be the only relevant criterion.

That is the reason why your code actually works. It is not as good a reason as I would wish for.

My plea: Please make the right hand sides of all those `#define`s simply `1`. You cannot save and restore more than the defined-ness status anyway, and that is all that matters. Similarly for those patterns in `include/math.h`, and the misleading comments there should be removed.

michaelld:
Just getting back to this after an extensive flu sickness. Good times ... not!

@ccorn I've worked out a better way to handle this I hope that avoids the issues you've documented. I'll do local testing & push them once I'm there. Hopefully we can get this next version dealt with before another set of changes come through here that require me to redo this PR yet again.

kencu:
agreed - let's get this done. I won't push anything else until it's finished.

michaelld:
Thx & I'll try to get it done, tested, and pushed today.

include/math.h
```diff
@@ -0,0 +1,104 @@
...
+ * wants <cmath>, they should #include that specific header.
+ */
+
+/* store prior values for _GLIBCXX_MATH_H and _GLIBCXX_INCLUDE_NEXT_C_HEADERS */
+
+#ifdef L_GLIBCXX_MATH_H
+#undef L_GLIBCXX_MATH_H
+#endif
+#ifdef _GLIBCXX_MATH_H
+#define L_GLIBCXX_MATH_H _GLIBCXX_MATH_H
```
michaelld: Aw crap I forgot about these !@#$ #ifdef's. ... let me fix them too like I did with those in <cmath> here ...

michaelld:
OK so I need to fixup the #ifdef's that are borked in <math.h> herein to match what I did in <cmath> ... but, anyone have issue what what I did in <cmath> in this regard? These changes do still work for me for example when building cmake on 10.5 using GCC5 or GCC6.

kencu:
Even a simple program that calls std::log2 will tell you if it’s fixed or broken, I found.

michaelld:
Well I did do "make test_cmath" first & verified that it worked. But, I wanted to also test on a port that I knew would otherwise fail.

Anyway: Wondering what you & @ccorn think of the new #ifdef's in <cmath> ...

kencu:
With gcc on Tiger, and with clang on all systems, as you noted, sometimes the builds succeed where they should fail. But I don't consider this a fault -- in fact, perhaps easier if the builds succeed, even if they technically shouldn't, given we're after ease of use and maximum compatibility rather than strict POSIX adherence. Otherwise, it seems committable.

kencu:
OK, I'll put out a new *-devel version and we'll see if we can break this out in the wild.

Thank you, Michael, for sticking with this through all the shenanigans.

michaelld:
Thanks @kencu ... I'll fixup the math.h #ifdef's to be cleaner & do another PR. Once that's finalized, we'll need to do a new release to get the new features into the main legacy-support port for all ports to use. Then I can update the cmake Portfile with the legacy PG. Speaking of which, with this latest cmath fix, "make test" for the whole of cmake does almost as well as it does on 10.1[1-4] ... just a few tests don't pass and 3 of those are because the test is faulty & I don't understand why the others fail but don't have time to deal with them:

> 99% tests passed, 6 tests failed out of 538
> 
> Label Time Summary:
> Label1    =   0.62 sec*proc (1 test)
> Label2    =   0.62 sec*proc (1 test)
> 
> Total Test time (real) = 11262.51 sec
> 
> The following tests FAILED:
>          80 - Architecture (Failed)
>         133 - CPackComponentsForAll-DragNDrop-default (Failed)
>         193 - BundleTest (Failed)
>         241 - CTestTestCrash (Failed)
>         378 - RunCMake.WriteBasicConfigVersionFile (Failed)
>         448 - RunCMake.FindGTK2 (Failed)

kencu:
One thing is the c++11 issue. Up to now, in my mind, this bug was not related to c++11, but was just a generic bug in Apple's headers that went back to 10.4. 

And I never really understood why std::log2 etc worked with gcc6 on 10.4 but not 10.5 and 10.6.

michaelld:
Looking at the default system <math.h>, the`llround` (etc) prototypes are protected because they are `long long` which isn't available on really old compilers (maybe GCC3? or GCC4.0?; GCC4.2 works correctly in my testing). I don't see that as a "bug", but then Apple didn't fix up modern compilers to set the correct macros to get these functions until OSX 10.7 I think -- which is definitely an oversight if nothing else. Apple also didn't fix up the protection macros to handle when in c++ mode.

Meanwhile the GCC folks messed up the macros in <cmath> to not correctly get these prototypes (when in c++ mode, which is what <cmath> is used for anyway duh).

The problem manifests itself in c++11 specifically because the standard (finally) defines `std::llround` (etc), so the C prototypes are required when using c++11 or more recent. This is not an issue in the c++ standard; it's a result of the c++ standard being correct and the API being buggy on older OSX and some (not even so much older GCC).

So, it's sort of a 1-2 bug/feature messup that is forced to be fixed with modern c++, and really should just be fixed for any reasonable GCC and any Clang. That's what we're trying to do here.

kencu:
I assume you already saw these guys, where we worked on this a couple of years ago, FWIW:

<https://trac.macports.org/ticket/53226>

<https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79017>

michaelld:
I hadn't see those tickets / issues / reports. The new <cmath> will address the MP ticket. The GCC folks fixed the issue I think in GCC8 rather than GCC7, but whatever I did the testing / verification & set the correct version in the macros used to determine whether to use this fix or not. Should work now for all GCC versions that are supposed to support std::log2 .

kencu:
great. Somewhere they backported the gcc fixes into gcc7 -- but not official releases of gcc6 or less (@riccardo did backport all these fixes into gcc48, gcc5, and gcc6, but we didn't integrate those yet).

OF interest, the same math errors occur when building with clang against the gcc headers (using -stdlib=macports-libstdc++). Hopefully this fix covers off that situation too.

michaelld:
hmmm ... not sure this fix as currently written will work with clang using gcc headers ... thinking the macros for GCC version will be too old ... but, who knows? Should be fixable to work with the correct macro selection. Let me see if I have clang installed on my PPC or 10.6 Intel computer to test.

kencu:
You won't have a usable on PPC unless you've been following along my magic clang-on-PPC project :> Soon -- soon.

PR #17, opened 2019-03-01
fixup! math.h

michaelld:
* correctly pass through to the next math.h when not required for legacy support;
* also fixup the macros to work like those in <cmath> herein.

### DIFF
--- a/include/math.h
+++ b/include/math.h
@@ -1,4 +1,3 @@
-
 /*
  * Copyright (c) 2010 Chris Jones <jonesc@macports.org>
  * Copyright (c) 2019 Michael Dickens <michaelld@macports.org>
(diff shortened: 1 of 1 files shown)


michaelld:
These changes work for me & make sense too to me. Once we get them merged, please create a new tagged release & update both legacy-support and legacy-support-devel accordingly. Then I can update cmake to use the cmath changes of recent.

kencu:
Given the big changes, and the fact that we haven't really built any other ports against this math change, could we have a few weeks to test this in *-devel?

michaelld:
Yes sure take time to do testing. I've tested on my various systems & these seem to work correctly when required & do nothing otherwise. Anyway, when the new cmath is formally added to legacy-support I'll update cmake to use it & then it'll build cleanly on PPC OSX & some older Intel OSX too.

kencu:
As far as I can tell this correctly works on all compilers on 10.4 through 10.6. I haven't been able to find a broken case yet.

You do have to keep in mind that gcc started defaulting to -std=c++11 on gcc6, and clang does not default to that ever (clang-7.0 defaults to -std=c++17, the previous versions to -std=c++03). And that if you're going to use -std=c++11 with clang then you also need to have a c++11 capable stdlib selected --- but if you do all that correctly, these builds all seem to work.

PR #19, opened 2019-03-16
Fix for OSX10.11 builds using Xcode8

cjones051073:
Possible fix for 

https://trac.macports.org/ticket/58199

### DIFF
--- a/include/time.h
+++ b/include/time.h
@@ -27,20 +27,11 @@
 /* Legacy implementation of clock_gettime */
 #if __MP_LEGACY_SUPPORT_GETTIME__
 
-#ifndef CLOCK_REALTIME
-/* These values are choosen to match use in macOS10.12+ */
-#define CLOCK_REALTIME             0
-#define CLOCK_MONOTONIC            6
-/* we do not implement these (yet). */
-/* #define CLOCK_MONOTONIC_RAW        4 */
-/* #define CLOCK_MONOTONIC_RAW_APPROX 5 */
-#endif
+/* One define types and methods if not already defined. */
+#if !defined(CLOCK_REALTIME) && !defined(CLOCK_MONOTONIC)
 
-/*
- * Define a type for the above CLOCK_* values; many OSs use 'enum' for
- * the values & then typedef the enum to 'clockid_t'; use 'int' for
- * easy compatibility.
- */
+#define CLOCK_REALTIME  0
+#define CLOCK_MONOTONIC 6
 typedef int clockid_t;
 
 #ifdef __cplusplus
@@ -53,6 +44,8 @@ extern int clock_getres ( clockid_t clk_id, struct timespec *ts );
 #ifdef __cplusplus
 }
 #endif
+
+#endif /* !defined(CLOCK_REALTIME) && !defined(CLOCK_MONOTONIC) */
 
 #endif /* _MP_LEGACY_SUPPORT_GETTIME__ */
 


michaelld:
Interesting idea. I have no idea if a duplicate (but otherwise identical) enum will create errors or warnings for a c/c++ compiler. I'll check this out on Monday when I'm at work (where my OSX build boxes are located).

cjones051073:
Nor do I. Above builds fine on 10.6, Fwiw.

cjones051073:
B.t.w. My thinking was if its just that on 10.11 the clockid_t enum exists, but the two methods do not, then that will be easy to fix by just suppressing the enum definition in our header for that OS.

michaelld:
Clang doesn't allow for enum redefinitions, or at least the Clang 8.2.1 provided in 10.11 ; here's the CMake build error log in this case:
```
/opt/local/include/LegacySupport/time.h:40:1: error: redefinition of enumerator '_CLOCK_REALTIME'
_CLOCK_REALTIME = 0,
^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include/time.h:153:1: note: previous definition is here
_CLOCK_REALTIME __CLOCK_AVAILABILITY = 0,
^
```

michaelld:
This isn't the correct solution here, unfortunately.

cjones051073:
It might not fix the 10.11 issue, but i could still make the case for it being the right thing to do anyway, as it means the legacy method matches more closely what is in 10.12+

cjones051073:
Bth, for me the simpliest quick fix for the issue in the ticket could well be ‘dont use xcode 8 on 10.11’

michaelld:
Agree on this change ... reopening!

michaelld:
We can't force people to not use Xcode 8.2.1 ... that's just not realistic.

michaelld:
BTW> I'm going to push a temporary fix to get folks with this issue going again. We can work out a better way down the road & then remove the temporary fix.

cjones051073:
@michaelld Could you try the following ?  Protect the definition of the enum in the time.h file we ship with

```
#ifndef __CLOCK_AVAILABILITY
enum .....
#endif
```

On my 10.11 Xcode 7 VM the above is not defined, so the legacy-support enum would be set.

cjones051073:
b.t.w. I've push to this branch the addition of the define check

michaelld:
OK that's a good idea. I'll check it out tomorrow morning.

cjones051073:
sadly, on a second look I don't think this is going to work, as the system time.h has at the end

```
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include/time.h:#undef __CLOCK_AVAILABILITY
```

so the __CLOCK_AVAILABILITY macro never makes it out of the internals of the header....

michaelld:
Hmmm ... yeah that's an issue. Let me take a look & see what I can come up with here. I think you're onto something with the 2-part checking: the first part for whether the OSX version meets the criteria, and the 2nd part looking for some specific macro that's defined only when `clock_gettime` is available in the SDK. My temporary fix in the cmake Portfile is at least taking care of the folks with the issue ... gives us some time / wiggle room to figure out a better way to do this!

cjones051073:
Yeah, exactly. Some macro that defines the xcode version, or sdk, being used, would be perfect. Not found one yet though...

michaelld:
-Maybe-: use `CLOCK_REALTIME` ... it works with Xcode 8.2.1, which provides `clock_gettime` and related code. I'll try it on 10.10 with Xcode 7.2.1 & see if it works there, where `clock_gettime` (etc) should not be defined.

michaelld:
Yes that macro is looking like an option. It has to protect all of the legacy added code, not just the enum. I'll push the changes shortly, after one more test locally.

cjones051073:
I beat you to it ;)

Not sure why I did not think of those macros myself. Guess I got distracted trying to understand Apples macros for OSX versions... god thats a mess.

michaelld:
Hmm .. let me test this latest change. I think there'll be an issue with the function redefinitions. I was literally just about to force push my fixes when you did yours LOL. Competition!

cjones051073:
yeah, I was not sure if the function definitions also needed to be hidden inside the check or not...

cjones051073:
just for reference, in my 10.11 Xcode 8 VM whether or not these macros exist depend on if you look a the `/usr/local` or Xcode SDK versions of `time.h`

```
MacVM1011xc8 ~ > grep -R CLOCK_REALTIME /usr/include
MacVM1011xc8 ~ > grep -R clockid_t /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include/time.h:} clockid_t;
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include/time.h:int clock_getres(clockid_t __clock_id, struct timespec *__res);
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include/time.h:int clock_gettime(clockid_t __clock_id, struct timespec *__tp);
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include/time.h:__uint64_t clock_gettime_nsec_np(clockid_t __clock_id);
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include/time.h:int clock_settime(clockid_t __clock_id, const struct timespec *__tp);
```

what this means in terms of whether or not the system library function is found at link time I am not sure..

michaelld:
Interesting! I see the same on my 10.11 box that has Xcode 8.2.1. /usr/include has the "old" headers while /Applications/Xcode.... has the new headers. Our fix using the CLOCK_REALTIME macro should work with either; no idea why CMake is picking the Xcode headers rather than the system headers ... whatever CMake!

michaelld:
OK so this version works for me nicely on 10.10 / Xcode 7 and 10.11 / Xcode 8. Please give it a read-through & try on your setup(s).

cjones051073:
Can you try and avoid the need for

```
#include <sys/cdefs.h>
```

we have had a lot of trouble with these wrapped headers including anything other than the `include_next` header they are wrapping, such that we aim to avoid including anything else now.

cjones051073:
If I understand if correct you are using `sys/cdefs.h` just for `__BEGIN_DECS`
etc. Can you please remove this and revert back to using

```
#ifdef __cplusplus
extern "C" {
#endif
```

etc. There was a good reason why we did this in the first place ;)

cjones051073:
It looks like the only real difference to what I had previously pushed (other than the DECL stuff) is that you also put the declaration of the two methods inside the new check as well, whereas I left them out.

cjones051073:
OK, I have push back my version, with the one change you had. Can you please give this specific version a try ?

michaelld:
Sure I'll give it a go when I'm back at work tomorrow where my OSX build machines are located. A quick read-through looks fine to me. I guess minimizing the need for system headers (e.g., DECL) is wise. I'll have to check math.h & see if I added them there; thinking so. Easy to swap out as you note.

include/time.h
```diff
@@ -54,6 +48,8 @@ extern int clock_getres ( clockid_t clk_id, struct timespec *ts );
 }
 #endif
 
+#endif /* !defined(CLOCK_REALTIME) && !defined(CLOCK_MONOTONIC) */za
```
michaelld: extra characters after the closing "*/"

cjones051073:
In the first instance we also included other 'system' headers in the wrapped headers, as well as the one it was specifically wrapping. It worked OK, most of the time, but then we ran into some very strange issues because of it, such as infinite include loops and cases where the wrong headers where being used. So now we aim to not include anything other than the one header they are wrapping (and the MP one, but this is fine as nothing else includes this).

include/time.h
cjones051073: thanks... emacs thick fingers... no fixed.

michaelld:
Yeah not using system headers makes sense. I'll fixup legacy's math.h to remove the DECL stuff.

michaelld:
All looks OK now reading again. I'll verify tomorrow on 10.10 and 10.11.

cjones051073:
I've just remember the main reason why we stopped including other system headers...

The issue was if we do this, then it can in third party builds, which then use our wrapper versions of those headers, change the order the headers get included in those builds. Now, in principle the headers as supposed to be safe to be reordered..... in practise though it does not always work out like that and we start to see build failures because of this implicit reordering. So now, we aim to not include anything else so that the original order is by construction preserved.

cjones051073:
My 10.11 Xcode 8 VM just built cmake just fine (with the legacy PortGroup put back in) using the version here.

cjones051073:
I've reverted back to using a plain int for clockid_t typedef, as the enum seemed to cause issue on 10.6.

michaelld:
Your latest change builds for me on 10.9, 10.10, and 10.11 ... my 10.5 PPC box is slow but it's looking positive there too. So LGTM for merging & we'll clean it up later if need be.

cjones051073:
OK thanks. I think I will merge this and push out a `legacy-support-devel` update with it, together with #21 .

PR #20, opened 2019-03-19
Miscellaneous cleanup

Ionic:
Whitespace canonicalization and rework of `README.md`, now featuring a true table, which hopefully is cleaner.

Wasn't able to use GitHub's markdown feature for the table, because it doesn't support row- or colspan.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-
-# MacPorts Support for legacy OSX versions
+# MacPorts Support for Legacy OSX Versions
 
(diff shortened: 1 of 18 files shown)


cjones051073:
I do like your changes to README.md, I agree that is cleaner to view. I also guess the clean up to the white space changes is a good idea. So in all LGTM.

cjones051073:
I know @kencu is on vacation, and as merging this will help with reviewing #21 I am going to merge this now. @kencu hope this is eventually OK with you...

PR #21, opened 2019-03-19
Add `fstatat64` function.

Ionic:
This function is considered heavily private, but was added as a syscall anyway.

It's not actually exposed in system header files, but applications could call it via `syscall()`.

I do not expect a lot of applications to make use of it.

This PR depends upon #20.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,8 @@ Wrapped headers are:
   </tr>
   <tr>
     <td rowspan="2"><code>sys/stat.h</code></td>
-    <td>Adds <code>fchmodat</code>, <code>fstatat</code>, and <code>mkdirat</code> functions</td>
+    <td>Adds <code>fchmodat</code>, <code>fstatat</code>, <code>fstatat64</code> (if required),
+        and <code>mkdirat</code> functions</td>
     <td>OSX10.9</td>
   </tr>
   <tr>
--- a/include/sys/stat.h
+++ b/include/sys/stat.h
@@ -31,6 +31,9 @@ extern "C" {
 #endif
   extern int fchmodat(int dirfd, const char *pathname, mode_t mode, int flags);
   extern int fstatat(int dirfd, const char *pathname, struct stat *buf, int flags);
+#if !__DARWIN_ONLY_64_BIT_INO_T
+  extern int fstatat64(int dirfd, const char *pathname, struct stat64 *buf, int flags);
+#endif
   extern int mkdirat(int dirfd, const char *pathname, mode_t mode);
 #ifdef __cplusplus
 }
--- a/src/atcalls.c
+++ b/src/atcalls.c
@@ -87,6 +87,9 @@ int unlinkat(int dirfd, const char *pathname, int flags);
 
 int fchmodat(int dirfd, const char *pathname, mode_t mode, int flags);
 int fstatat(int dirfd, const char *pathname, struct stat *buf, int flags);
+#if !__DARWIN_ONLY_64_BIT_INO_T
+int fstatat64(int dirfd, const char *pathname, struct stat64 *buf, int flags);
+#endif
 int mkdirat(int dirfd, const char *pathname, mode_t mode);
 
 /* #include <sys/fcntl.h> */
@@ -236,6 +239,18 @@ int fstatat(int dirfd, const char *pathname, struct stat *buf, int flags)
         return ATCALL(dirfd, pathname, stat(pathname, buf));
     }
 }
+
+#if !__DARWIN_ONLY_64_BIT_INO_T
+int fstatat64(int dirfd, const char *pathname, struct stat64 *buf, int flags)
+{
+    ERR_ON(EINVAL, flags & ~AT_SYMLINK_NOFOLLOW);
+    if (flags & AT_SYMLINK_NOFOLLOW) {
+        return ATCALL(dirfd, pathname, lstat64(pathname, buf));
+    } else {
+        return ATCALL(dirfd, pathname, stat64(pathname, buf));
+    }
+}
+#endif
 
 int getattrlistat(int dirfd, const char *pathname, struct attrlist *a,
                                  void *buf, size_t size, unsigned long flags)
--- a/test/test_fstatat64.c
+++ b/test/test_fstatat64.c
@@ -1,0 +1,258 @@
+/*
+ * Copyright (c) 2019 Mihai Moldovan <ionic@ionic.de>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include <sys/stat.h>
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <unistd.h>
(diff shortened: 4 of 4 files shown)


cjones051073:
Thanks, but please remove the whitespace changes, so it is easier to see what you are actually proposing. The whitespace changes can be the subject of a separate PR, if you think they are required.

cjones051073:
So sorry... I negated to look at PR 20, which is the whitespace changes... to fast to comment..

Ionic:
Heh, yeah, I split it up correctly but also rebased this branch against the other to make merging easier eventually (even witthout needing a special merge commit). Sadly GitHub is not "smart" enough to detect inter-PR dependencies.

Ionic:
Had to move one whitespace change that I forgot to stage correctly to the other PR, hence the last update.

cjones051073:
I have merged #20 but now this PR has conflicts. Can you please can a look @Ionic ?

Ionic:
Rebased.

Do you know when he'll be back? I'd like to have a release with these changes in so that I can depend upon the new release version in `go`.

cjones051073:
@kencu is not the only one who does releases ;)

Ionic:
Well, sure, in theory, anyone of us could do it, including me, but it's mostly his baby...

cjones051073:
Could you please add a simple test application to the PR to test the new feature ?

cjones051073:
i.e. add something that runs when 'make test' is run.

Ionic:
Agree'd, good idea. I'll look into it.

cjones051073:
> Well, sure, in theory, anyone of us could do it, including me, but it's mostly his baby...

I have also made my share of releases.

Note though, I would probably suggest before making a tagged release of this pushing this out as the `legacy-support-devel` port, which is designed for eactly this sort of thing, to test changes before they hit the main repo. Once I am happy there is nothing here that is going to impact current features, I will be happy to merge it myself and update the devel port. You will then just need to manually migrate from `legact-support` to `legacy-support-devel` to test.

cjones051073:
Your changes appear to be generating compiler warnings.

```
cc -c -Iinclude -Os -Wall -fPIC src/atcalls.c -o src/atcalls.dl.o
src/atcalls.c: In function ‘fstatat64’:
src/atcalls.c:248: warning: ‘lstat64’ is deprecated (declared at /usr/include/sys/stat.h:464)
src/atcalls.c:248: warning: ‘lstat64’ is deprecated (declared at /usr/include/sys/stat.h:464)
src/atcalls.c:250: warning: ‘stat64’ is deprecated (declared at /usr/include/sys/stat.h:465)
src/atcalls.c:250: warning: ‘stat64’ is deprecated (declared at /usr/include/sys/stat.h:465)
```

This is on 10.6, so even on this old OS these calls are already depreciated. Do we really want to be adding something that is using these ?

Ionic:
Added a testing application just now. Not sure how useful it really is because I'm essentially only comparing the binary data of a struct obtained via `stat64` and `fstatat64` - with and without different CWD's.

The problem with `legacy-support-devel` is that the `legacysupport` PG hard-depends upon the `legacy-support` port. I may rewrite this into a path-based dependency which can be satisfied by either port. Since I need this feature for the `go` port, though, I'll have to wait for a release anyway.

In any case, all this adds is yet another function that is not available otherwise, so there's really no way how it could affect any other part.

> Your changes appear to be generating compiler warnings.

Yeah, that's by design. I may not have explained it well in my original post, so let me have another go.

`struct stat64` and any functions with a `64` postfix (like `stat64()`, `lstat64()`, ...) are deprecated features - deprecated since 10.6. Back in the very old days (no idea what OS version), there was only a `stat` struct and it had a 32-bit inode entry.  At some point (OS X 10.5), Apple noticed that this was not good enough any longer and they created a new struct called `stat64` with a 64-bit inode number and a few other changes, complete with functions postfixed `64` that worked with this new struct. To retain ABI compatibility, they left the original functions unmodified, but added another set of functions which would be generated with an `$INODE64` suffix in their **symbol names** and would work on the newer struct. Whether code that uses plain `stat()` calls would the 64-bit variant or the 32-bit variant was dependent on the OS version. 10.5 used the 32-bit variant by default. You may have noticed that we both have functions with `$INODE64` in their symbol name and another set of functions with a `64` postfix... if this sounds redundant, well, it is. So, with 10.6, the default implementation was changed to the 64-bit variant and the `64`-postfixed functions declared deprecated. That's leaving the non-deprecated `$INODE64` symbol magic for ABI compatibility and everything's fine. Even newer OS versions will eventually drop the `stat64` struct completely - or rather rename it to `struct`. I don't know if there's any OS version that does that yet, but we've got safe guards anyway, should that happen. On such platforms, neither the declaration nor the implementation will be available, so everything's good.

The weird part is `fstatat64`. This function was only introduced with 10.10 - together with its counterpart `fstatat` - LONG AFTER the `64`-postfixed functions were declared deprecated. Why? No idea, but Apple did that. And not just that: try to grep for `fstatat64` in `/usr/include` on a 10.10+ machine. The only result you'll come up with is a syscall define. So it's not even a properly declared function, but only available via use of `syscall()`. The kernel (XNU) includes code to handle it just fine.

I think that the reason for this mess is Linux-compatibility. It does sound weird, but Linux is also offering an `fstatat64` syscall - and at that the only OS I could find. BSD-derivates seem to have switched to 64-bit inodes long ago in a hard-cut way, so they don't need any of that - which is surprising.


Since `go` wraps all syscalls, they generate some code from the definitions found in the Apple-provided header files and `fstatat64` suddenly made an appearance when this code was updated on a 10.10 machine. From what I can tell it's not even used in the core `go` code. Patching it out is difficult, though, because, like I said, it's part of generated files, so the generator would have to be modified to mask out some syscalls.


In any case, there is a chance that other applications also use `fstatat64()` (even if just because it's there), so I think it has deserved a place in `legacy-support`.

cjones051073:
> The problem with `legacy-support-devel` is that the `legacysupport` PG hard-depends upon the `legacy-support` port. I may rewrite this into a path-based dependency which can be satisfied by either port. Since I need this feature for the `go` port, though, I'll have to wait for a release anyway.

No it doesn't... See 

https://github.com/macports/macports-ports/blob/master/_resources/port1.0/group/legacysupport-1.0.tcl

line 29

Ionic:
Err, yeah, now that I looked at it again... I confused it with [line 48](https://github.com/macports/macports-ports/blob/master/_resources/port1.0/group/legacysupport-1.0.tcl#L48)

cjones051073:
OK... Thanks for the detailed description above, very helpful.

If the compiler warnings are expected then that is fine. Perhaps it would though be nice to then suppress them, by passing the appropriate `-Wno-....` flag to the `atcalls.c` compilation, but that is a minor detail.

As it stands then I am happy to merge this, and then push out a `legacy-support-devel` update that uses it (together with #19 I think). As I expain above the legacy support PortGroup already supports using this port instead of ``legacy-support` so that should not be an issue for test purposes. I have used this myself a lot so know it works.

Before pushing this out though as a regular `legacy-support` release though I would prefer to wait for @kencu to take a look...

kencu:
Thanks for the contribution to this little project, and the great explanation. I read about 64 bit inode pointers, APFS, the legacy FS supports, etc this morning.... getting pretty nerdy, I guess :>

kencu:
Looks fine to me, ie I can't see how it might break anything on looking it over. Feel free to release away!

cjones051073:
Thanks @kencu.

@Ionic can you please try out the updated `legacy-support-devel` and report back confirming it works with your go ports. If that is the case, we can then look to making a full release.

Ionic:
I've already tried multiple times with the patches backported locally to `legacy-support`, but I also updated `legacy-support-devel` now and indeed `go` built successfully again.

Everything's fine on my side, I'll update the `go` port after seeing a new `legacy-support` release and after the environment stuff is fixed, or rather, after Joshua pushes port changes that show how to handle different base versions differently (macports/macports-base#116).

cjones051073:
If you also need changes to base to be made available, then you are going to have to wait until not only the above is merged, but also a new release is made of MacPorts base including those changes, and these do not happen so often, so you might be waiting a while.

mojca:
Just curious: on which OS version are you now able to build `go`?

Ionic:
The base changes are optional. The port would work without them, but requires another patch to work around the current limitations. After base is fixed, I can drop it conditionally and pass more complex environment variables.

`go` has been working on 10.9 for me now, I don't (easily) have a way to test older OS versions. I'll push my changes whenever the `legacy-support` release is around.

cjones051073:
OK I see. I'll make a full `legacy-support` release today then so you can then push your go changes. The build bots will then tell you how far back the builds work (unless they fail because of things like disk space issues, which sadly is increasingly common on some of the builders...)

Ionic:
Don't worry, this can all wait. :)

I have other stuff to do before I can push any stuff, so it'll likely take me another few days if not a week. And I have to give the port's maintainer a heads-up anyway by uploading a patch for review.

cjones051073:
No problem. Making a release with hithub is pretty trivial, jut a few clicks here and there...

cjones051073:
Its done.

PR #24, opened 2019-12-07
Create fdopendir variations (including ground work for multiple variations support)

Ionic:
This PR included support for building multiple variations of specific symbols.

Currently, it only does that for `fdopendir`. One consumer of this are newer versions of `go`, but you already know that.

We will probably need to eventually do this for other `atcalls` as well, but I'm too lazy to check which ones are necessary and refactor (split out) the code to do that. The good part is that this operation should be pretty easy to do, since this PR lays a solid foundation for this sort of things.

Also, documentation is added for this new feature (plus a bit of more general documentation that was missing).

Once this is merged, I will update the `legacy-support-devel` port to the new build procedure.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,7 @@ LD              ?= ld
 LDFLAGS         ?=
 AR              ?= ar
 UNAME           ?= /usr/bin/uname
+SED             ?= /usr/bin/sed
 
 MKINSTALLDIRS    = install -d -m 755
 INSTALL_PROGRAM  = install -c -m 755
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,7 @@ AR              ?= ar
 UNAME           ?= /usr/bin/uname
 SED             ?= /usr/bin/sed
 GREP            ?= /usr/bin/grep
+CP              ?= /bin/cp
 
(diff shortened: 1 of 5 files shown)


kencu:
Thanks for all the time you spent to factor this apart so we/I could follow it all more clearly. I appreciate the effort involved as well. It looks mergeable to me. I will download it on a few systems and run the test suite and see what happens.

The complexity goes up, naturally, with the added features.

I am also running into some headaches with libuv that needs different contructs for the _close command based on the system version (10.4 PPC, 10.5 PPC, and Intel 32bit/64bit all have different _close symbols to be called) so I can see that we might well have more use for this infrastructure once it is in place and working (and I actually understand how to use it!).

Ionic:
Might be the same issue as with `go`, but I really don't know what `libuv` does.

Sorry that it took me so long to actually properly PR this.

Ionic:
Argh, missed the notification here. Okay, will need to get the `-devel` port up to speed soon.

Ionic:
`-devel` subport updated. Check to see if you like it.

Once the software gets a proper release, we can merge the behavior together again (keeping only the newer behavior in the `Portfile`) and I can finally get the `go` port working again.

PR #25, opened 2020-02-10
which: exit 1 when the file is not found

miniupnp:
allows for constructs like
$ which tool || which toolalternative


kencu:
I tried this exact thing yesterday and it didn't seem to work -- I must have typo'd it or something. I'll give it another look!

cjones051073:
LGTM

kencu:
yes, works. thanks! We'll get Tiger working like Mojave sooner or later :>

PR #26, opened 2020-05-05
fsgetpath: implement replacement for 10.6+

kencu:
needed for systems prior to 10.13
two methods of implementing fsgetpath are  available
we'll use the simpler syscall method for now
other method is included in source, protected by blocker
not simple to write a test for this at present

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,11 @@ Wrapped headers are:
     <td>OSX10.9</td>
   </tr>
   <tr>
+    <td><code>sys/fsgetpath.h</code></td>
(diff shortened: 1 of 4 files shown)


kencu:
@cjones051073 -- hey Chris, if you have a moment to cast eyes on this. It is needed to fix "searchfs" on older systems.

cjones051073:
Seems reasonable enough, although I cannot really comment on if the syscall implementation does the right thing or not as I am not familiar with the method being emulated here. I understand it might not be easy in this case but a test application would be very useful...

PR #27, opened 2020-05-11
getattrlistat - Use void* instead of struct attrlist*

cjones051073:
Triggered by discussion at 

https://github.com/macports/macports-ports/commit/78b2cb450f7f56f0d1e6f66c44cd6e3a3fd742b1

Scanning through the API for `getatrlistat` in various OSX versions it turns out the API is slightly different to what we have here. Basically one argument is `void*` and not a specific type.  e.g. from macOS (10.15)

```
int     getattrlistat(int, const char *, void *, void *, size_t, unsigned long) __OSX_AVAILABLE_STARTING(__MAC_10_10, __IPHONE_8_0);
```

This MR just updates the API to match.

@kencu FYI

### DIFF
--- a/include/sys/unistd.h
+++ b/include/sys/unistd.h
@@ -53,7 +53,7 @@ typedef __darwin_gid_t		gid_t;
 
 __MP__BEGIN_DECLS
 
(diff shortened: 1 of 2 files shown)


kencu:
looks reasonable, thanks

PR #29, opened 2020-08-27
fix fd leaks in fdopendir

kzc:
`fdopendir` is supposed to take control of its `dirfd` parameter upon success.
An unrelated edge case in error handling was also fixed to not leak memory and a fd.

See https://github.com/evanw/esbuild/issues/348 for context.

### DIFF
--- a/src/fdopendir.c
+++ b/src/fdopendir.c
@@ -30,6 +30,13 @@ int best_fchdir(int dirfd);
 
 #define PROTECT_ERRNO(what)  ({ int __err = (errno); what; errno = __err; })
 
+/*
+ * Implementation behavior largely follows these man page descriptions:
+ *
+ * https://www.freebsd.org/cgi/man.cgi?query=fdopendir&sektion=3
+ * https://linux.die.net/man/3/fdopendir
+ */
+
 DIR *fdopendir(int dirfd) {
     DIR *dir;
     struct stat st;
@@ -43,8 +50,12 @@ DIR *fdopendir(int dirfd) {
         return 0;
     }
 
-    if (dirfd == AT_FDCWD)
-        return opendir (".");
+    if (dirfd == AT_FDCWD) {
+        dir = opendir (".");
+        /* dirfd can be closed only upon success */
+        if (dir) PROTECT_ERRNO(close(dirfd));
+        return dir;
+    }
 
     oldCWD = open(".", O_RDONLY);
     if (oldCWD == -1)
@@ -58,12 +69,16 @@ DIR *fdopendir(int dirfd) {
     dir = opendir (".");
 
     if (best_fchdir(oldCWD) < 0) {
+        if (dir) PROTECT_ERRNO(closedir(dir));
         if (oldCWD != -1) PROTECT_ERRNO(close(oldCWD));
         return 0;
     }
 
     if (oldCWD != -1)
         PROTECT_ERRNO(close(oldCWD));
+
+    /* dirfd can be closed only upon success */
+    if (dir && dirfd != -1) PROTECT_ERRNO(close(dirfd));
 
     return dir;
 }


kzc:
Somewhat on topic... I have no idea how macports works, but I see the port health of Go is red for 10.9 and prior OSX versions. I was able to get Go and any program built with it to work on OSX 10.9.5 using the instructions in https://github.com/evanw/esbuild/issues/81#issuecomment-680147925.

ryandesign:
> I was able to get Go and any program built with it to work on OSX 10.9.5 using the instructions in [evanw/esbuild#81 (comment)](https://github.com/evanw/esbuild/issues/81#issuecomment-680147925).

Thanks, I've added a link to your comment to https://trac.macports.org/ticket/58935#comment:33 which is where we are tracking the go failure on 10.9 and earlier.

src/fdopendir.c
```diff
@@ -43,8 +55,12 @@ DIR *fdopendir(int dirfd) {
         return 0;
     }
 
-    if (dirfd == AT_FDCWD)
-        return opendir (".");
+    if (dirfd == AT_FDCWD) {
+        dir = opendir (".");
+        /* dirfd can be closed only upon success */
+        if (dir && dirfd != -1) PROTECT_ERRNO(close(dirfd));
```
michaelld, line 61: We knows here that `dirfd != -1` (since it equals `AT_FDCWD`, which is -2 according to `sys/fcntl.h`), so that part can be removed.

kzc:
> go can be built on 10.9 using legacy support and DYLD_INSERT_LIBRARIES

I updated my comment in the esbuild thread to reflect that the `go` toolchain was not built - the official golang release was untarred and used as-is.

src/fdopendir.c
```diff
@@ -30,6 +30,18 @@ int best_fchdir(int dirfd);
...
+
+  Upon successful return from fdopendir(), the file descriptor is under the
+  control of the system, and if any attempt is made to close the file
+  descriptor, or to modify the state of the associated description other
+  than by means of closedir(), readdir(), readdir_r(), or rewinddir(), the
+  behavior is undefined.  Upon calling closedir() the file descriptor is
+  closed.  The FD_CLOEXEC flag is set on the file descriptor by a
+  successful call to fdopendir().
+*/
+
```
michaelld, line 44: Please also note https://linux.die.net/man/3/fdopendir  :
```
After a successful call to fdopendir(), fd is used internally by the implementation, and should not otherwise be used by the application.

The fdopendir() function leaves the setting of the close-on-exec flag unchanged for the file descriptor, fd. POSIX.1-200x leaves it unspecified whether a successful call to fdopendir() will set the close-on-exec flag for the file descriptor, fd.
```

src/fdopendir.c
kzc: done

src/fdopendir.c
kzc: I replaced the documentation with the two man page links instead to avoid possible licensing issues with inlining the text.

michaelld: nice!

kzc:
@kencu Any concerns with this PR? It allows the official go 1.15 release and any program created with it to work correctly on OSX 10.9.5 in file handle constrained environments.

kencu:
Sorry, I didn't see this before.

I'll have to look it over and try it out on 10.4 to whereever it tops out to see what happens. In the meantime, you have made these changes on your own system I presume so you have that working for you.

kencu:
looks reasonable, thank you for the time you spent improving our legacy library. We have hopes that this will prove ever more useful for those of us with older Apple systems that are still useful hardware-wise, if abandoned software-wise by upstream.

kzc:
Thanks to everyone in macports who make this great software.

> Getting go to support the C bindings on systems where upstream has dumped support is tricky, though. If you have skills in this area, much appreciate your help. Specifically, making go link in accessory libraries like legacy-support is a supreme PITA.

Would `install_name_tool` help in this regard?

* https://trac.macports.org/ticket/29838
* https://discourse.nixos.org/t/patchelf-for-mach-o-shared-libraries/562/3

PR #32, opened 2021-01-29
include: support undefined wrapping macros.

Ionic:
This just fixes noisy build warnings, no change in behavior intended.

I could have just defined the macros to 0 if undefined instead in `include/MacportsLegacySupport.h`, but that way is probably more generic.

### DIFF
--- a/include/MacportsLegacySupport.h
+++ b/include/MacportsLegacySupport.h
@@ -92,8 +92,8 @@
 #define __MP_LEGACY_SUPPORT_REALPATH_WRAP__   (__APPLE__ && __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1060)
 
 /*  realpath() wrap has bail-out macros in case we want to disable only function wrapping */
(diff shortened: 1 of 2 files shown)


kencu:
I just have to test these things out on Tiger and Leopard, where we have had a lot of trouble keeping all these older compilers happy (like gcc-4.0).

The other thing that happens is that people define them to 0 to turn things off. But because they are "defined" then they pass the test, instead of fail.

SO that turns out to be bad, in the end.

Ionic:
> I just have to test these things out on Tiger and Leopard, where we have had a lot of trouble keeping all these older compilers happy (like gcc-4.0).

Standard C, but go ahead.

> The other thing that happens is that people define them to 0 to turn things off. But because they are "defined" then they pass the test, instead of fail.

I hope you meant to `1`, not `0`. `0` disables the disabling of wrapping, so... the wrappers get enabled. That's the default and the logic of `!defined (foo) || !foo` should be sound.

Generally, I wouldn't use negations in the first place. It's a common practice to avoid them due to double-negation reasons. It's just confusing. But... oh well. No over-engineering, I guess.

PR #33, opened 2021-01-30
src: split out shared functionality to common-priv.h.

Ionic:
Less code duplication.

This header file lives in `src/` because it's not meant to be installed. It's fully private.

While that may not sound like a big deal yet, we'll use this more extensively soon. Trust me, it's a good thing.

As an added benefit, we can also consolidate `__pthread_fchdir()` and `best_fchdir`, which were exactly the same. Yay.

### DIFF
--- a/src/common-priv.h
+++ b/src/common-priv.h
@@ -1,0 +1,77 @@
+/*-
+ * Copyright (c) 2019-2021 Ken Cunningham kencu@macports.org
+ *
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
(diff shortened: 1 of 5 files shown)


kencu:
pthread_fchdir doesn’t exist on tiger so that’s why we had best_fchdir.

If someone removed that when I wasn’t looking, that’s probably why we had all those recent errors on Tiger I couldn’t explain. 

I’ll have to put it back.

Ionic:
Rebased against current master.

No, nobody removed it. But we've had two functions doing exactly the same -- `best_fchdir` (in its own compilation unit `src/best_fchdir.c`) and `__pthread_fchdir` (local to `src/atcalls.c`). I consolidated both to use `best_fchdir` instead and removed the duplicated one in `src/atcalls.c`.

Ionic:
Hm... actually, the defines are getting complicated, let's redo this a bit.

Ionic:
That looks way better/less complicated and fixes a bug in my implementation that I haven't realized before due to added consistency when it comes to when the `ATCALL` macros or `best_fchdir` are actually needed.

Ionic:
And then, because I cannot do without typos, a typo fix.

src/common-priv.h
```diff
@@ -0,0 +1,77 @@
+/*-
+ * Copyright (c) 2019 Ken Cunningham kencu@macports.org
```
michaelld, line 2: Need to update this (C) ... either yourself or MP & the new year

src/common-priv.h
```diff
@@ -0,0 +1,77 @@
...
+#include <unistd.h>
+#endif /* __MP_LEGACY_SUPPORT_NEED_ATCALL_MACROS__ */
+
+#define PROTECT_ERRNO(what)  ({ int __err = (errno); what; errno = __err; })
+#define ERR_ON(code, what)   if (what) { errno = (code); return -1; }
+
+#ifndef SYS___pthread_fchdir
+# define SYS___pthread_fchdir 349
+#endif
+
```
michaelld, line 40: extra \n to be removed

michaelld (changes requested):
A couple minor nit-picks ... otherwise LGTM & makes sense to provide this functionality in a common but private header

src/common-priv.h
Ionic: I intentionally left it as-is, mostly because this is just a refactor/move of code without actual changes.

But yeah, the minor changes in preprocessor conditions could justify an update, so I'll do that.

src/common-priv.h
Ionic: That, likewise, is intentional. I like to group logical things via double newlines and find that the visual cues make reading code a bit easier. Personal preference, though.

If I remove it, I'd better also remove the other double newline for consistency.

src/common-priv.h
michaelld: You are creating a new file, which needs a new (C): either totally new or keep the old (C) for the code as copied, but update it for anything new.

src/common-priv.h
michaelld: Hmmm ... I don't think we're consistent in this spacing. How about we just leave it as you've done it, since that's less work :)

src/common-priv.h
Ionic: Updated, with interval to this year.

michaelld: works for me; thx!

michaelld (approved):
LGTM

michaelld:
I'll leave this up to @kencu now ... I think it's a good change!

kencu (approved):

PR #35, opened 2021-02-06
MacportsLegacySupport.h: avoid warnings due to undefined behavior.

Ionic:
The defined unary preprocessor operator explicitly exhibits undefined
behavior when not used in conditional inclusion (i.e., preprocessor
directives that are not #if or #elif).

Compilers (correctly) warn for that, so avoid such usage.

On the plus side, avoiding it is really easy with object-like macros.

This draft is supposed to come after #34, so disabled for now.

### DIFF
--- a/include/MacportsLegacySupport.h
+++ b/include/MacportsLegacySupport.h
@@ -92,9 +92,19 @@
 #define __MP_LEGACY_SUPPORT_REALPATH_WRAP__   (__APPLE__ && __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1060)
 
 /*  realpath() wrap has bail-out macros in case we want to disable only function wrapping */
-#define __ENABLE_MP_LEGACY_SUPPORT_REALPATH_WRAP__  ((!defined (__DISABLE_MP_LEGACY_SUPPORT_FUNCTION_WRAPPING__) || !__DISABLE_MP_LEGACY_SUPPORT_FUNCTION_WRAPPING__)  && \
-						     (!defined (__DISABLE_MP_LEGACY_SUPPORT_REALPATH_WRAP__) || !__DISABLE_MP_LEGACY_SUPPORT_REALPATH_WRAP__)          && \
-						     __MP_LEGACY_SUPPORT_REALPATH_WRAP__)
+#if ((!defined (__DISABLE_MP_LEGACY_SUPPORT_FUNCTION_WRAPPING__) || !__DISABLE_MP_LEGACY_SUPPORT_FUNCTION_WRAPPING__)  && \
+     (!defined (__DISABLE_MP_LEGACY_SUPPORT_REALPATH_WRAP__) || !__DISABLE_MP_LEGACY_SUPPORT_REALPATH_WRAP__)          && \
+     __MP_LEGACY_SUPPORT_REALPATH_WRAP__)
+/*
+ * Do NOT try to "optimize" this by replacing the if preprocessor construct with a
+ * single macro definition.
+ *
+ * This invokes undefined behavior ([cpp.cond]p4/§6.10.1.4) and leads to ugly warnings.
+ */
+#define __ENABLE_MP_LEGACY_SUPPORT_REALPATH_WRAP__  1
+#else
+#define __ENABLE_MP_LEGACY_SUPPORT_REALPATH_WRAP__  0
+#endif
 
 /* lsmod does not exist on Tiger */
 #define __MP_LEGACY_SUPPORT_LSMOD__           (__APPLE__ && __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1050)
@@ -109,9 +119,14 @@
 #define __MP_LEGACY_SUPPORT_SYSCONF_WRAP__    (__APPLE__ && __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 101100)
 
 /*  sysconf() wrap has bail-out macros in case we want to disable only function wrapping */
-#define __ENABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__  ((!defined (__DISABLE_MP_LEGACY_SUPPORT_FUNCTION_WRAPPING__) || !__DISABLE_MP_LEGACY_SUPPORT_FUNCTION_WRAPPING__)  && \
-						    (!defined (__DISABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__) || !__DISABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__)            && \
-						    __MP_LEGACY_SUPPORT_SYSCONF_WRAP__)
+#if ((!defined (__DISABLE_MP_LEGACY_SUPPORT_FUNCTION_WRAPPING__) || !__DISABLE_MP_LEGACY_SUPPORT_FUNCTION_WRAPPING__)  && \
+     (!defined (__DISABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__) || !__DISABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__)            && \
+     __MP_LEGACY_SUPPORT_SYSCONF_WRAP__)
+/* Likewise, do NOT try to condense this into one single macro definition. */
+#define __ENABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__  1
+#else
+#define __ENABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__  0
+#endif
 
 /* pthread_rwlock_initializer is not defined on Tiger */
 #define __MP_LEGACY_SUPPORT_PTHREAD_RWLOCK__  (__APPLE__ && __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1050)
--- a/src/macports_legacy_realpath.c
+++ b/src/macports_legacy_realpath.c
@@ -14,15 +14,20 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
+/*
+ * We need this blocker so as to not get caught in our own wrap.
+ *
+ * Importantly, do this BEFORE including our feature header, since it can only be
+ * included once and subsequent changes won't have an effect.
+ */
+#undef __DISABLE_MP_LEGACY_SUPPORT_REALPATH_WRAP__
+#define __DISABLE_MP_LEGACY_SUPPORT_REALPATH_WRAP__ 1
+
 /* MP support header */
 #include "MacportsLegacySupport.h"
 
 /* realpath wrap */
 #if __MP_LEGACY_SUPPORT_REALPATH_WRAP__
-
-/* we need this blocker so as to not get caught in our own wrap */
-#undef __DISABLE_MP_LEGACY_SUPPORT_REALPATH_WRAP__
-#define __DISABLE_MP_LEGACY_SUPPORT_REALPATH_WRAP__ 1
 
 #include <limits.h>
 #include <stdlib.h>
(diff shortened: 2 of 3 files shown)


Ionic:
Oh, wow, this leads to the sysconf test segfaulting, so a definite DO NOT MERGE currently.

kencu:
I think we may have to consider reverting your previous commit until I can see just what warnings you were seeing there, as i never saw any warnings from the legacysupport headers on any systems before, and we can't have them putting out any warning (as stated before) due to that causing -Werror failures in ports.

Ionic:
I'll fix the warnings that got added with this PR, but I currently have to figure out what is causing the new, different, faulty behavior introduced with this commit. Let me get this done first, please.

kencu:
Just a FYI -- the entire mechanism that @ccorn came up with for overriding functions, by relabelling the ```__asm```s to new functions, no longer works with newer clang compilers.

gcc still allows relabelling asms, but clang, as of a certain commit a couple of versions ago, does not. 

So I will need to come up with an entirely new method of doing this (probably running 'fixincludes' on the headers in the SDK and moving them into legacysupport/include, -- OR -- patch clang to again allow relabelling asms.

There were plenty of complaints about clang shutting that behaviour down (apparently it broke building the linux kernel, or perhaps it was libc, when they did that) -- but they were adamant about it, and would not revert the commit that did that.

Whether this has anything to do with the segfaults you might be seeing remains to be seen, but if that is related, then it needs a deeper fix than just playing with header defines. 

Just so you don't waste all day on it, if that is it.

kencu:
Thanks.

This is quite a different animal now, as you are setting the defines in a completely different place.

The first go around, we tested this for weeks on every system from 10.4 up, with many compilers, to make sure it worked.

I don't know how long it might take me to sort out if this works properly. I accept that it no longer segfaults on your system but I need to make sure it works to do what it did before.

You know, for now, if you are wanting a working ghc/cabal/stack for 10.9 that doesn't have the utimensat issue, just get the ones from my github repo that work on 10.6 up -- .

Ionic:
Nope, it was something different. There was an actual bug in the code, even without my changes.

The implementation files can't define a macro affecting the feature file `MacportsLegacySupport.h` AFTER including it, since a secondary include won't change anything (due to its include guard).

Moved it up and now it behaves like it should. User applications remain unaffected by that change.

---

As to why that my original PR was needed: I guess that the warnings I saw only came up when some software used `-Wundef`.

Ionic:
I'll keep this PR as a draft until `utimensat` is in, then rebase one last time and undraft it.

Ionic:
Yeah, the segfault was actually caused by a stack overflow since the `sysconf` function called itself recursively, without ever calling the "real" one.

If the assembler labeling stops working, it'll need to be redone anyway in a different manner, but... I guess we'll have to tackle this when it gets relevant. You definitely don't want to keep patching clang to accept the relabeling mechanism - it'll get difficult very quickly in such a fast-moving project with a lot of changes.

It shouldn't be difficult to get rid of the asm relabelling, but it was convenient and aligned well with Apple's own code. I wonder how they are handling this in newer releases that will use a compiler not supporting such relabeling. I certainly don't believe they will patch their CLT clang versions to support it.

kencu:
There was no bug. That is not how that file worked.

kencu:
Would you, by any chance, be able to show me the warnings that led to all this? I'm not overly looking forward to all the time I am going to have to spend to sort this all out - for warnings I never saw.

kencu:
The assem relabing fails now -- there are a bunch of tickets about it.

I have been carrying patches for clang for years to make it do all kinds of things we want it to do.

Ionic:
Like I said, I don't quite remember what software I hit this on. It might have been `qemu`, I'd have to recheck to know that for sure.

I think that qemu is quite pedantic on warnings and potential bugs, so they turned on `-Wundef`, which lead to those warnings and caused me to look into it (disabled my commit temporarily on my local repo to show the old code's "issue"):

```
%  cat test/test_warn.c
#include <unistd.h>

int main (const int argc, const char * const * argv) {
  return (0);
}
% CFLAGS="-Wundef" CXXFLAGS="-Wundef" make check
[...]
cc -c -std=c99 -Iinclude -Wundef test/test_warn.c -o test/test_warn.o
In file included from test/test_warn.c:1:
In file included from include/unistd.h:22:
In file included from include/MacportsLegacySupport.h:24:
include/MacportsLegacyWrappers/wrapper_macros.h:27:6: warning: '__DISABLE_MP_LEGACY_SUPPORT_FUNCTION_WRAPPING__' is not defined, evaluates to 0 [-Wundef]
#if !__DISABLE_MP_LEGACY_SUPPORT_FUNCTION_WRAPPING__
     ^
In file included from test/test_warn.c:1:
include/unistd.h:24:5: warning: '__DISABLE_MP_LEGACY_SUPPORT_FUNCTION_WRAPPING__' is not defined, evaluates to 0 [-Wundef]
#if __ENABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__
    ^
include/MacportsLegacySupport.h:112:54: note: expanded from macro '__ENABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__'
#define __ENABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__  (!__DISABLE_MP_LEGACY_SUPPORT_FUNCTION_WRAPPING__  && \
                                                     ^
In file included from test/test_warn.c:1:
include/unistd.h:24:5: warning: '__DISABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__' is not defined, evaluates to 0 [-Wundef]
include/MacportsLegacySupport.h:113:12: note: expanded from macro '__ENABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__'
                                                    !__DISABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__       && \
                                                     ^
In file included from test/test_warn.c:1:
include/unistd.h:34:5: warning: '__DISABLE_MP_LEGACY_SUPPORT_FUNCTION_WRAPPING__' is not defined, evaluates to 0 [-Wundef]
#if __ENABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__
    ^
include/MacportsLegacySupport.h:112:54: note: expanded from macro '__ENABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__'
#define __ENABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__  (!__DISABLE_MP_LEGACY_SUPPORT_FUNCTION_WRAPPING__  && \
                                                     ^
In file included from test/test_warn.c:1:
include/unistd.h:34:5: warning: '__DISABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__' is not defined, evaluates to 0 [-Wundef]
include/MacportsLegacySupport.h:113:12: note: expanded from macro '__ENABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__'
                                                    !__DISABLE_MP_LEGACY_SUPPORT_SYSCONF_WRAP__       && \
                                                     ^
5 warnings generated.
cc -Llib  test/test_warn.o -lMacportsLegacySupport -o test/test_warn
test/test_warn
cc -c -std=c99 -Iinclude -Wundef test/test_wcscasecmp.c -o test/test_wcscasecmp.o
In file included from test/test_wcscasecmp.c:18:
In file included from include/wchar.h:22:
In file included from include/MacportsLegacySupport.h:24:
include/MacportsLegacyWrappers/wrapper_macros.h:27:6: warning: '__DISABLE_MP_LEGACY_SUPPORT_FUNCTION_WRAPPING__' is not defined, evaluates to 0 [-Wundef]
#if !__DISABLE_MP_LEGACY_SUPPORT_FUNCTION_WRAPPING__
     ^
1 warning generated.
```

Ionic:
If you want, I can look into the asm relabelling stuff and rewrite that into macro interception. Should be pretty easy and work everywhere. Let's just do that. I'll start working on it and create another draft PR.

Ionic:
Ugh, I spoke to soon. Simple macro interception won't work.

Function interposing might work, but I wonder if that will continue to be the case. Worth a try, I guess.

kencu:
Yes, I tried a number of versions of macro interception but unreliable. I tried function interposing as well, but it is not transparent enough.

Redefining asm labels works (IIRC) with all gcc versions, and with clang up to about clang-5.0, possibly clang-6.0, not clang-7.0+.


But -- what I think **will** work is just rewriting the appropriate headers on Tiger, Leopard, and possibly a few newer systems to use the new function names instead of the old ones, right in the header file. Then just call those preferentially as we do now, using legacysupport's include overrides.

@michaelld and I have been hashing out a "fixincludes" type mechanism where the old header is copied in from the SDK in "/" and then patched to use the new "legacysupport" function name.

And -- if none of that works smoothly enough -- fix clang by allowing redefining asm labels again, at least on sytems were this matters.


See these links for clues:

https://trac.macports.org/ticket/58677

https://trac.macports.org/ticket/59621

https://reviews.llvm.org/D16171

https://github.com/llvm/llvm-project/commit/6ee53bc0850c80287a6c5905af0ffd2101513021

kencu:
The tiny tiny tiny patch to clang to allow it to define asms looks really minor.... I think I will have to go with that approach on affected systems, at least to try it out:

<https://trac.macports.org/ticket/58677#comment:37>

kencu:
Ionic -- my deepest apologies. I reverted the previous commit to get us back to the "last working state", as it has been (basically) trouble-free for many years now.

I see your skills here at fixing up these things, and will work on getting these warnings fully suppressed.

Perhaps we might just use the ```#pragma``` that tells the compiler it's a system header and to shuttup about the warnings... that is apparently automatically done if you include legacysupport using -isystem.

But we'll get it. I just have to keep it legible enough to not collapse in obfuscation so my tiny little mind knows which way is up and which way is down when I go to fix some broken thing.

Ionic:
Well... we're back to square one now. With the revert, we'll get warnings for builds when using `-Wundef`.

This PR would have fixed it once and for all. The segfault issue was fixed quite some time ago.

I've only waiting for #34 to land so that I could have rebased this PR and have it merged too.

Well, no harm done, really. I can re-introduce the changes here now again without causing warnings.

Though, speaking of warnings: there are lot of deprecation warnings on newer systems due to the use of `stat64` as well. I wonder why no one ever ran into those with `-Wall`. I guess I'll have to fix them eventually as well, but not here and right now.

---

Regarding the unrelated `asm` relabeling stuff: I've been toying with that two weeks ago, but couldn't come up with a solution that universally works. It either worked for linked libraries (by just providing the symbols verbatim, shadowing the version in the system libraries) or working when using `DYLD_INSERT_LIBRARIES` via the function interposing feature. At some point I considered building the dylib twice - once for normal linking purposes and once for preloading purposes. Like... `libMacportsLegacySupport-preload.dylib`. I haven't had time to get this done properly, though, and probably won't find any for the next few weeks due to work.

kencu:
legacysupport has survived so far by being well-vetted and reliable. There are no bugs in the headers at present that I know of. And really, with hundreds of ports using it now, it's not trivial to start changing things to fix one warning that bugs someone without seeing what it might do to many other ports that were working with it a certain way.

If we want to stop a warning that really bugs you, we can probably just add ```#pragma GCC system_header```  without changing anything else. <https://gcc.gnu.org/onlinedocs/cpp/System-Headers.html> (although I haven't tried it yet.)

Our current method of overriding functions works most of the time now, except for a couple of niche cases on 10.5 that I am trying to sort out when using clang. I am not thinking of doing a whole-hog different method of doing legacysupport completely.

Most likely, I will just rewrite the stdlib.h on 10.5 to use the new function name directly. Or tweak clang to allow asm renaming again.

kencu:
I'm going to close this one so it doesn't accidentally get committed by some well-meaning soul, and will work on something else over time.

Ionic:
Yep, I'll just keep the patches locally for myself.

kencu:
we've been down these roads before https://github.com/macports/macports-legacy-support/commit/73ce23ea710f8277bd6b495a2d6a8e76873c6091 and for various reasons things always had to be removed.

SO -- there's a long history of butt-bites with these things ;> 

I appreciate the try -- and I think there is (most likely) a simpler way to fix these warnings without #define spaghetti by using

```
#pragma GCC system_header
```

as above. Just have to get around to actually trying it.

PR #39, opened 2021-09-18
which: Point to MacPorts-provided tcsh

evanmiller:
I occasionally get an error running "which" on Tiger because the system-provided `csh` doesn't like my `$LS_COLORS`:

```
$ which which
csh: Unknown colorls variable `rs'.
```

Modern tcsh provided by MP does not appear to have this issue – so I propose switching to that.

If this PR is approved, the `legacy-support` Portfile will need a run-time `tcsh` dependency added on Tiger.

### DIFF
--- a/tiger_only/bin/which
+++ b/tiger_only/bin/which
@@ -1,4 +1,4 @@
-#!/bin/csh
+#!/opt/local/bin/tcsh
 #
 # DO NOT USE "csh -f"
 #


kencu:
I don't think we will do this one, Evan.

I think we are going to stick with our very-slightly-modified "which" that you will find is already installed by legacysupport into /opt/local/bin/which. There are just too many ways things can go wrong otherwise. The which command is embedded into a number of build scripts, and right now it works very much like he newer systems "which". It has stood the test of time.

Of course, you are completely welcome to do anything you like on your own systems, and I would encourage you to do so.

thanks, keep 'em coming, but not this one...

evanmiller:
/opt/local/bin/which intermittently fails for me, including in build scripts, because `$LS_COLORS` trips up /bin/csh before the rest of `which` can even run. This is a fairly annoying and took me a while to track down why my builds were bugging out. I will try again with a narrow, no-dependency PR and see what you think.

PR #40, opened 2021-09-19
which: Fix error when LS_COLORS is set

evanmiller:
A much narrower solution than that proposed in #39

### DIFF
--- a/tiger_only/bin/which
+++ b/tiger_only/bin/which
@@ -1,4 +1,7 @@
-#!/bin/csh
+#!/usr/bin/env LS_COLORS="" /bin/csh
+#
+# The above shebang line fixes intermittent errors like:
+# csh: Unknown colorls variable `rs'.
 #
 # DO NOT USE "csh -f"
 #


kencu:
yep, OK.

kencu:
do you think LS_COLORS is something that very many Tiger users would set in the environment? I don't think I ever heard of it before, all these years and years of using Tiger.

And if we're going to special case a fix for LS_COLORS, goodness knows what other env vars might someone want to set in their environment that they will also want special-cased? So in general, I don't think we should move to these kinds of things. If someone wants to set an env var for their use, and it messes with the Tiger setup we've so carefully cobbled together, well -- they can figure out how to work around it too, I would say. It's pretty easy, after all. copy the "which" script to a place of your pleasure, edit it to your liking, and stick it on the front of your PATH. 

Done.

evanmiller:
I'll need to dig into it but something else is setting my LS_COLORS - xterm, fish shell, or something like that.

Tiger's `csh` can't even start if it doesn't like the `LS_COLORS` value. To me this is a bug in `/bin/csh` and everything that depends on `/bin/csh`, including all the MP builds that use the `which` script.

If there are other innocuous environmental variables that completely prevent Tiger's `/bin/csh` from starting, and thus MP builds from operating as expected, then yes I would say add them to the list here.

evanmiller:
I'm thinking the setter of `LS_COLORS` is fish shell... This only became compilable / usable again on Tiger a couple months ago (e.g. https://github.com/fish-shell/fish-shell/pull/8097) so maybe that's why I'm the only one reporting this issue so far.

kencu:
but Evan, my point is we can't start special-casing things in the system tools to work around weird bugs in fish, etc. It's too risky, and the wrong way to manage these issues.

Please don't PR any more, and I'm inclined to even revert this one.

evanmiller:
It's not a bug in fish; it's a bug in `csh` that was fixed in 2009:

https://lists.opensuse.org/archives/list/bugs@lists.opensuse.org/message/2PJGX6E62EUM44FQF54KS2GB3CZWSEXH/

What would you consider the right way to manage a bug in a system-provided shell like this? It likely affects Leopard as well, maybe Snow Leopard too – so actually for that reason I'm sympathetic to a reversion here.

LS_COLORS can be managed via `gdircolors`, which is provided by MP `coreutils`. So this affects any shell where a user has done any `ls` color customization – not just fish.

I don't mind fixing this locally but it seems rather burdensome for the average MP Terminal user to diagnose the issue, figure out the culprit, and add their own layer of workaround scripts.

Happy to continue this conversation on a new Trac ticket if that would be more appropriate.

kencu:
you can certainly feel free to open a ticket with your suggestion that macports should replace the system shells in older systems with newer versions, if that is your gist.

There are tickets about this already I recall, as we have discussed this many times in the past, and many of us have in fact replaced bash or others with newer, statically-linked versions to get past known vulnerabilities like heartbleed.

but these are manual steps people can do. The average terminal user uses Terminal.app, does not use fish on Tiger, and does not touch the default colours. So there is no burden.

what we support is a stock system running stock tools as supplied by Apple. A minor change in "which" to work like 10.5+ (because build scripts were not accounting for the Tiger behaviour) actually took quite a bit of testing to get right. Our first three or four versions still showed errors at times with some software builds, and it was over 6 months before it was exactly right.

You might believe your suggested changes here to support colour shells using fish etc could not possibly have any deleterious effect, and I'm just going to say that you would be very surprised at how fragile things truly are.

PR #43, opened 2021-11-11
Define `u_char` etc on Tiger using Leopard rules

evanmiller:
A number of ports run into trouble compiling on Tiger because `_XOPEN_SOURCE` results in `_POSIX_C_SOURCE` being defined, thus hiding the `u_char`, `u_int`, and other typedefs.

Leopard and later provide a workaround in the form of `_DARWIN_C_SOURCE`; follow this behavior.

See: https://trac.macports.org/ticket/52070
See: https://trac.macports.org/ticket/62997

(Note that this PR will do nothing in practical terms until the `legacysupport` Portfile is modified.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,11 @@ Wrapped headers and replaced functions are:
     <td>OSX10.4</td>
   </tr>
   <tr>
+    <td><code>sys/types.h</code></td>
+    <td>Adds definitions for <code>u_char</code>, <code>u_short</code>, <code>u_int</code>, <code>u_long</code>, <code>ushort</code>, and <code>uint</code> types that can be exposed via <code>_DARWIN_C_SOURCE</code></td>
+    <td>OSX10.4</td>
+  </tr>
+  <tr>
     <td rowspan="3"><code>sys/unistd.h</code></td>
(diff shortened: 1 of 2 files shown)


michaelld:
Please let us know when this PR is out of draft form. Right now the new header won't be installed & thus this must still be a draft PR ... yes?

cjones051073:
The code in the port file need to be improved. There is no reason to have to explicitly list the headers there to be copie over for tiger - A filesystem glob should be used to find whats available under ${worksrcpath}/tiger_only/include and just install the lot...

cjones051073:
https://github.com/macports/macports-ports/commit/23b3f213fb9feab549cdf68eb6b318e8939beb77

cjones051073:
Changes look fine to me. As they can only affect tiger builds, which few can test, will merge and make a new `legacy-support-devel` with them.

cjones051073:
https://github.com/macports/macports-ports/commit/662a0e7442d01115de6743e0172eadd9bb111611

please try it out

evanmiller:
@cjones051073 The latest commits seem to install Tiger-only headers OK, thanks!

PR #44, opened 2021-11-11
fcntl.h: Expose flock(2) and friends on Tiger

evanmiller:
Match the 10.5 SDK with use of `_DARWIN_C_SOURCE`.

There will be a few more PRs like this in the coming days as I try to get `tmux` building on 10.4. This is a difficult issue to patch in application code; the usual workaround requires something like

```
#ifdef _POSIX_C_SOURCE
#define OLD_POSIX_C_SOURCE _POSIX_C_SOURCE
#undef _POSIX_C_SOURCE
#include <fcntl.h>
#define _POSIX_C_SOURCE OLD_POSIX_C_SOURCE
#else
#include <fcntl.h>
#endif
```

For a similar (but not directly impacted) issue see https://trac.macports.org/ticket/53421


evanmiller:
Converting these to Draft while I try a different tactic with tmux.

evanmiller:
I ended up using a different strategy for compiling tmux, and so I am closing these PRs.

PR #52, opened 2023-01-22
clock_gettime: Cache timebase

krackers:
On versions of osx before 10.12, mach_timebase_info is a direct syscall (see dyld-239.3/glue.c#L434) which is order of magnitudes slower than the subsequent mach_absolute_time() call. Hence it's better to just cache it after the first call.

Since 10.12+, Apple also realized this and began caching mach_timebase_info calls at the libc layer (https://opensource.apple.com/source/xnu/xnu-3789.41.3/libsyscall/wrappers/mach_timebase_info.c.auto.html)

### DIFF
--- a/src/time.c
+++ b/src/time.c
@@ -31,7 +31,7 @@ int clock_gettime( clockid_t clk_id, struct timespec *ts )
   int ret = -1;
   if ( ts )
   {
-    if      ( CLOCK_REALTIME == clk_id )
+    if ( CLOCK_REALTIME == clk_id )
     {
       struct timeval tv;
       ret = gettimeofday(&tv, NULL);
@@ -40,10 +40,12 @@ int clock_gettime( clockid_t clk_id, struct timespec *ts )
     }
     else if ( CLOCK_MONOTONIC == clk_id || CLOCK_MONOTONIC_RAW == clk_id )
     {
-      const uint64_t clock = mach_absolute_time();
-      mach_timebase_info_data_t timebase;
-      mach_timebase_info(&timebase);
-      uint64_t tdiff = clock * timebase.numer / timebase.denom;
+      const uint64_t t = mach_absolute_time();
+      static mach_timebase_info_data_t timebase;
+      if (timebase.numer == 0 || timebase.denom == 0) {
+        mach_timebase_info(&timebase);
+      }
+      const uint64_t tdiff = t * timebase.numer / timebase.denom;
       if ( CLOCK_MONOTONIC == clk_id ) {
         tdiff = THOUSAND * ( tdiff / THOUSAND );
       }


mascguy:
This looks like a great improvement!

Added a few folks to review/comment.

cjones051073:
The implementation here doesn't look thread safe. Its unclear to me whether we need it to be which is why I didn't cache anything in the first place (I was well aware of this).

jmroot:
As the comment in the linked Apple source code says,
```
	/*
	 * This is racy, but because it is safe to initialize twice we avoid a
	 * barrier in the fast path by risking double initialization.
	 */
```

src/time.c
```diff
@@ -40,13 +40,12 @@ int clock_gettime( clockid_t clk_id, struct timespec *ts )
     }
     else if ( CLOCK_MONOTONIC == clk_id || CLOCK_MONOTONIC_RAW == clk_id )
     {
-      const uint64_t clock = mach_absolute_time();
-      mach_timebase_info_data_t timebase;
-      mach_timebase_info(&timebase);
-      uint64_t tdiff = clock * timebase.numer / timebase.denom;
-      if ( CLOCK_MONOTONIC == clk_id ) {
-        tdiff = THOUSAND * ( tdiff / THOUSAND );
-      }
```
mascguy, line 49: Do we still need the check for `CLOCK_MONOTONIC`, along with the adjustment if true?

mascguy: It looks like the rounding logic was added last month, via the following commit:

https://github.com/macports/macports-legacy-support/commit/88852aa8928004f1b02e40d80d924ccf8fc31195

@cjones051073 Chris, your thoughts?

cjones051073: So please put back the scaling for `CLOCK_MONOTONIC` you appear to have removed,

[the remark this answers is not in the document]
mascguy: > Please restore these lines.

No worries, thanks for confirming that they're needed!

mascguy (approved):
Squashed the commits, and built locally. Looks good!

mascguy:
Thanks so much for your contribution @krackers!

fhgwright:
Jeez, cache the numerator and denominator separately, preserving the exact rational nature.

mascguy:
> Jeez, cache the numerator and denominator separately, preserving the exact rational nature.

The logic is identical to what Apple is using in later macOS releases, and is battle-tested.

Why use anything else?

fhgwright:
The Apple code linked above doesn't show the actual time calculation, but it does show the numerator and denominator being cached separately, rather than cacheing the _approximate_ ratio, which is what the code here is doing.

It shouldn't be necessary to explain that `counter * (numerator / denominator) != (counter * numerator) / denominator`, especially in fixed-point arithmetic.

mascguy:
> The Apple code linked above doesn't show the actual time calculation, but it does show the numerator and denominator being cached separately, rather than cacheing the _approximate_ ratio, which is what the code here is doing.
> 
> It shouldn't be necessary to explain that `counter * (numerator / denominator) != (counter * numerator) / denominator`, especially in fixed-point arithmetic.

You're welcome to submit a PR to change it, if you feel that strongly about it...

cjones051073:
> The Apple code linked above doesn't show the actual time calculation, but it does show the numerator and denominator being cached separately, rather than cacheing the _approximate_ ratio, which is what the code here is doing.
> 
> It shouldn't be necessary to explain that `counter * (numerator / denominator) != (counter * numerator) / denominator`, especially in fixed-point arithmetic.

@fhgwright Your analysis is incorrect. The code here does *not* cache the result of the ratio calculation, but an instance of the `mach_timebase_info_data_t` struct, which internally saves the denominator and numerator separately. 

```
static mach_timebase_info_data_t timebase;
if (timebase.numer == 0 || timebase.denom == 0) {
     mach_timebase_info(&timebase);
}
```

This is exactly the same as what Apple caches.

The ratio is then calculated for each call to the method, to scale the absolute time for that call

```
const uint64_t t = mach_absolute_time();
uint64_t tdiff = t * timebase.numer / timebase.denom;
```

fhgwright:
@ cjones051073 
> @fhgwright Your analysis is incorrect. The code here does _not_ cache the result of the ratio calculation, but an instance of the `mach_timebase_info_data_t` struct, which internally saves the denominator and numerator separately.

Sorry.  I didn't look carefully enough before commenting.  Though I guess that was also true of your comment regarding thread safety. :-)

cjones051073:
> @ cjones051073
> 
> > @fhgwright Your analysis is incorrect. The code here does _not_ cache the result of the ratio calculation, but an instance of the `mach_timebase_info_data_t` struct, which internally saves the denominator and numerator separately.
> 
> Sorry. I didn't look carefully enough before commenting. Though I guess that was also true of your comment regarding thread safety. :-)

No, my comment there is correct. The code is inherently not thread safe, two threads could initialise the static cache at the same time. The comment in Apple’s implementation shows they are aware of this but consider it acceptable in this case.

PR #59, opened 2023-06-30
stdio.h: add open_memstream (OS X < 10.13)

sideeffect42:
To fix the build of [uacme](https://github.com/ndilieto/uacme) on older macOS versions (I guess < 10.13) an implementation of `open_memstream` is needed.

This implementation originates from https://piumarta.com/software/memstream/.

Maybe a backport of the function from Apple's libc would've been possible, but I didn't know where to find it, so I took the first one which looked reasonable.


mascguy:
Great, this is something I had planned to do anyway. But glad someone else beat me to it!

See: https://trac.macports.org/ticket/63785#comment:5

mascguy (approved):
LGTM, I'll create a new release of `legacy-support-devel` to test in the wild.

Thank you so much for your contribution!

PR #60, opened 2023-07-04
stdio.h: add fmemopen (OS X < 10.13)

mascguy:
### stdio.h: add fmemopen (OS X < 10.13)

- License: Apache-2
- Credit: https://github.com/tmm1/fmemopen
- Original Code (no longer available): https://github.com/ingenuitas/python-tesseract
- Test Code Credit: FreeBSD Project


mascguy:
While builds look fine for older macOS releases, the tests are crashing with a Segmentation Fault. I copied them verbatim from FreeBSD, so I'll compare against Apple's. (Theirs' appear to also be copied verbatim, but need to diff them to be sure.)

mascguy:
> While builds look fine for older macOS releases, the tests are crashing with a Segmentation Fault. I copied them verbatim from FreeBSD, so I'll compare against Apple's. (The latter also appear to be copied verbatim, but need to diff them to be sure.)

I'm glad I added the FreeBSD tests, as they've uncovered some limitations with the implementation I chose.

So next up is trying to use the FreeBSD incarnation, which is also what Apple uses anyway. Didn't consider that originally, but it makes more sense...

mascguy:
Updated to use the FreeBSD implementation, and all tests pass.

mascguy:
p.s. Josh updated the licenses for `legacy-support`, to include BSD. So I think (?) we're good there too:

https://github.com/macports/macports-ports/commit/6cc1b506cb3be841bd7a3300375b810a02694a7b

jmroot:
> p.s. Josh updated the licenses for `legacy-support`, to include BSD. So I think (?) we're good there too:

Yeah, there were already other BSD-licensed C files in the project.

PR #61, opened 2023-07-17
Fix fdopendir so dirfd stays open until closedir (for fstatat etc)

macportsraf:
This pull request changes `fdopendir()` so that the supplied directory file descriptor remains open until `closedir()` is called. This makes it possible for functions like `fstatat()` and `unlinkat()` to use the directory file descriptor (without changing its state) while traversing a directory hierarchy.

This was prompted by a pull request to get the sysutils/rawhide port working with legacysupport:

https://github.com/macports/macports-ports/pull/19047

https://trac.macports.org/ticket/67754

It changes `include/dirent.h` to include a wrapper struct and macro for `DIR` that contains the real `DIR*` as well as an `int` to store the supplied directory file descriptor. All of the functions that take a `DIR*` as an argument need wrappers as well.

There are wrapper macros for `readdir()`, `readdir_r()`, `rewinddir()`, `seekdir()` and `telldir()`, and there are wrapper functions and macros for `opendir()`, `closedir()` and `dirfd()`.

It changes `src/fdopendir.c` to change the implementation of `fdopendir()`. More on that below.

There is an additional source file (`src/fdopendir_support.c`). This was needed because the support functions (for `opendir()`, `closedir()` and `dirfd()`) couldn't reside in `src/fdopendir.c` because it uses `stat` and the wrappers don't. Compilation failed because `src/fdopendir.c` is compiled multiple times for different architectures but only a single function there was expected, and any additional functions end up being defined multiple times with the same name. The wrapper function for `dirfd()` only needed to be a function because the real `dirfd()` is already a macro.

There are two additional test files (`test/test_traverse.c` and `test/test_traverse_cwd.c`). They test two similar directory hierarchy traversals. One traverses a named directory. The other traverses the current directory (`"."`). Originally, they exhibited different errors (`Bad file descriptor` or `No such file or directory`) in subdirectories.

The `fdopendir()` implementation was changed to create and return the wrapper struct which stores the supplied file descriptor, so that it could remain open. Other functions would take the wrapper struct as an argument. When the `closedir()` wrapper is called, it closes the stored file descriptor.

That change was enough to fix uses of `fstatat()`, but a problem remained. The `test/test_traverse_cwd.c` test would `chdir()` into the test directory that it had created, then traverse `"."`, and then it would `chdir("..")` to get back to the starting directory to clean up the test directory that it had created, but the `chdir("..")` silently failed to work. It returned `0` but the current working directory was unchanged. It seemed impossible.

This was fixed in a way that might not be ideal. Replacing the two uses of `best_fchdir()` with `fchdir()` in `fdopendir()` fixed the problem, and the subsequent `chdir("..")` started working correctly.

According to `best_fchdir()`, it uses `__pthread_fchdir()` on macos-10.5 and later and it uses the real `fchdir()` on macos-10.4. I have no idea if `__pthread_fchdir()` really is the best `fchdir()`. I'm testing with macos-10.6. Perhaps the `__pthread_fchdir()` doesn't work properly on macos-10.6. Maybe it's something else.

Let me know if there is a better way to solve the `chdir()` problem. If `__pthread_fchdir` doesn't work on macos-10.6 but it does work on other systems, I could restore the calls to `best_fchdir()` if someone wants to run the tests on different systems to see if `test/test_traverse_cwd.c` works on any systems below macos-10.10. If some do, perhaps the implementation of `best_fchdir()` could be changed to only use `__pthread_fchdir()` on those systems, and use `fchdir()` on other systems. But `best_fchdir()` is used with atcalls and presumably works fine there. So maybe it doesn't need to change, and maybe it's OK to just not use it in `fdopendir()`.

P.S. There was a `FIXME` relating to tests failing on macos-10.4 when the supplied file descriptor is closed in `fdopendir()`, so it allowed the file descriptor leak to occur on macos-10.4. That can probably be considered fixed as well.

### DIFF
--- a/include/dirent.h
+++ b/include/dirent.h
@@ -1,6 +1,7 @@
 
 /*
  * Copyright (c) 2019
+ * Copyright (c) 2023 raf <raf@raf.org>, Tavian Barnes <tavianator@tavianator.com>
  *
  * Permission to use, copy, modify, and distribute this software for any
  * purpose with or without fee is hereby granted, provided that the above
@@ -27,6 +28,14 @@
 /* fdopendir */
 #if __MP_LEGACY_SUPPORT_FDOPENDIR__
 
+/* Wrapper struct for DIR */
+typedef struct __MP_LEGACY_SUPPORT_DIR __MP_LEGACY_SUPPORT_DIR;
+struct __MP_LEGACY_SUPPORT_DIR {
+    DIR *__mpls_dir;
+    int __mpls_dirfd;
+};
+#define DIR __MP_LEGACY_SUPPORT_DIR
+
 __MP__BEGIN_DECLS
 
 #ifndef __DARWIN_ALIAS_I
@@ -35,7 +44,47 @@ extern DIR *fdopendir(int fd) __DARWIN_ALIAS(fdopendir);
 extern DIR *fdopendir(int fd) __DARWIN_ALIAS_I(fdopendir);
 #endif
 
-__MP__END_DECLS
+/* Wrapper functions/macros to support fdopendir */
+extern DIR *__mpls_opendir(const char *name);
+extern int __mpls_closedir(DIR *dir);
+extern int __mpls_dirfd(DIR *dir);
+
+#define opendir(name)      __mpls_opendir(name)
+#define closedir(dir)      __mpls_closedir(dir)
+
+#ifndef __MP_LEGACY_SUPPORT_NO_DIRFD_MACRO
+#undef dirfd
+#define dirfd(dir)         __mpls_dirfd(dir)
 #endif
+
+static inline struct dirent *__mpls_readdir(DIR *dir) {
+    return readdir(dir->__mpls_dir);
+}
+
+static inline int __mpls_readdir_r(DIR *dir, struct dirent *entry, struct dirent **result) {
+    return readdir_r(dir->__mpls_dir, entry, result);
+}
+
+static inline void __mpls_rewinddir(DIR *dir) {
+    rewinddir(dir->__mpls_dir);
+}
+
+static inline void __mpls_seekdir(DIR *dir, long loc) {
+    seekdir(dir->__mpls_dir, loc);
+}
+
+static inline long __mpls_telldir(DIR *dir) {
+    return telldir(dir->__mpls_dir);
+}
+
+#define readdir __mpls_readdir
+#define readdir_r __mpls_readdir_r
+#define rewinddir __mpls_rewinddir
+#define seekdir __mpls_seekdir
+#define telldir __mpls_telldir
+
+__MP__END_DECLS
+
+#endif /* __MP_LEGACY_SUPPORT_FDOPENDIR__ */
 
 #endif /* _MACPORTS_DIRENT_H_ */
--- a/src/fdopendir.c
+++ b/src/fdopendir.c
@@ -1,6 +1,7 @@
 
-/*-
+/*
  * Copyright (c) 2019
+ * Copyright (c) 2023 raf <raf@raf.org>, Tavian Barnes <tavianator@tavianator.com>
  *
  *
  * THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
@@ -22,11 +23,15 @@
 
 #include "common-priv.h"
 
+#define __MP_LEGACY_SUPPORT_NO_DIRFD_MACRO
+
+#include <stdlib.h>
 #include <dirent.h>
 #include <sys/errno.h>
-#include <sys/stat.h>
-#include <unistd.h>
-#include <fcntl.h>
+
+#undef DIR
+#undef opendir
+#undef closedir
 
 
 /*
@@ -34,62 +39,107 @@
  *
  * https://www.freebsd.org/cgi/man.cgi?query=fdopendir&sektion=3
  * https://linux.die.net/man/3/fdopendir
+ *
+ * On success, this function returns allocated memory that must be
+ * deallocated by __mpls_closedir() (see closedir() macro in <dirent.h>).
  */
 
-DIR *fdopendir(int dirfd) {
-    DIR *dir;
-    struct stat st;
-    int oldCWD = -1;
+__MP_LEGACY_SUPPORT_DIR *fdopendir(int dirfd) {
 
-    if (fstat(dirfd, &st) < 0)
-        return 0;
+    /* Check dirfd here (for macos-10.4, see _ATCALL() and best_fchdir()) */
 
-    if (!S_ISDIR(st.st_mode)) {
-        errno = ENOTDIR;
+    if (dirfd != AT_FDCWD && dirfd < 0) {
+        errno = EBADF;
         return 0;
     }
 
-    if (dirfd == AT_FDCWD) {
-        dir = opendir (".");
-        /* dirfd can be closed only upon success */
-        if (dir) PROTECT_ERRNO(close(dirfd));
-        return dir;
-    }
+    /* Open the supplied directory safely */
 
-    oldCWD = open(".", O_RDONLY);
-    if (oldCWD == -1)
+    DIR *dir = _ATCALL(dirfd, ".", NULL, opendir("."));
+    if (!dir)
         return 0;
 
-    if(best_fchdir(dirfd) < 0) {
-        if (oldCWD != -1) PROTECT_ERRNO(close(oldCWD));
-        return 0;
-    }
+    /* Wrap it and return it (with the supplied directory file descriptor) */
 
-    dir = opendir (".");
-
-    if (best_fchdir(oldCWD) < 0) {
-        if (dir) PROTECT_ERRNO(closedir(dir));
-        if (oldCWD != -1) PROTECT_ERRNO(close(oldCWD));
+    __MP_LEGACY_SUPPORT_DIR *mplsdir = malloc(sizeof(*mplsdir));
+    if (!mplsdir) {
+        (void)closedir(dir);
+        errno = ENOMEM;
         return 0;
     }
 
-    if (oldCWD != -1)
-        PROTECT_ERRNO(close(oldCWD));
+    mplsdir->__mpls_dir = dir;
+    mplsdir->__mpls_dirfd = dirfd;
 
+    return mplsdir;
+}
 
-    /*
-     * FIXME -- this recently added bit makes the fdopendir tests fail on Tiger.
-     * check the whole commit where it was added to make sure it is
-     * doing the proper thing on all systems. Probably need more extensive tests
-     * to execise the whole system more aggressively.
-     */
+/*
+ * Wrapped version of opendir() for fdopendir() compatibility
+ *
+ * On success, this function returns allocated memory that must be
+ * deallocated by __mpls_closedir() (see closedir() macro in <dirent.h>).
+ */
 
-#if __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ >= 1050
-    /* dirfd can be closed only upon success */
-    if (dir && dirfd != -1) PROTECT_ERRNO(close(dirfd));
-#endif
+__MP_LEGACY_SUPPORT_DIR *__mpls_opendir(const char *name) {
 
-    return dir;
+    DIR *dir = opendir(name);
+    if (!dir)
+        return 0;
+
+    __MP_LEGACY_SUPPORT_DIR *mplsdir = malloc(sizeof(*mplsdir));
+    if (!mplsdir) {
+        (void)closedir(dir);
+        errno = ENOMEM;
+        return 0;
+    }
+
+    mplsdir->__mpls_dir = dir;
+    mplsdir->__mpls_dirfd = -1;
+
+    return mplsdir;
+}
+
+/*
+ * Wrapped version of closedir() for fdopendir() compatibility (see
+ * closedir() macro in <dirent.h>).
+ *
+ * This function deallocates memory that was allocated by fdopendir() or
+ * __mpls_opendir().
+ */
+
+int __mpls_closedir(__MP_LEGACY_SUPPORT_DIR *mplsdir) {
+
+    if (!mplsdir) {
+        errno = EBADF;
+        return -1;
+    }
+
+    int rc = closedir(mplsdir->__mpls_dir);
+
+    if (mplsdir->__mpls_dirfd != AT_FDCWD && mplsdir->__mpls_dirfd != -1)
+        PROTECT_ERRNO(close(mplsdir->__mpls_dirfd));
+
+    free(mplsdir);
+
+    return rc;
+}
+
+/*
+ * Wrapped version of dirfd() for fdopendir() compatibility because dirfd()
+ * itself is already a macro (see dirfd() macro in <dirent.h>).
+ */
+
+int __mpls_dirfd(__MP_LEGACY_SUPPORT_DIR *mplsdir) {
+
+    /* Return the supplied directory file descriptor if there was one */
+
+    if (mplsdir->__mpls_dirfd != AT_FDCWD && mplsdir->__mpls_dirfd != -1)
+        return mplsdir->__mpls_dirfd;
+
+    /* Otherwise call the underlying dirfd() macro */
+
+    return dirfd(mplsdir->__mpls_dir);
 }
 
 #endif /* __MP_LEGACY_SUPPORT_FDOPENDIR__ */
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ FORCE_ARCH      ?=
 ARCHFLAGS       ?=
 LIPO            ?= lipo
 CC              ?= cc $(ARCHFLAGS)
-CFLAGS          ?= -Os -Wall
+CFLAGS          ?= -Os -Wall -Wno-deprecated-declarations
 DLIBCFLAGS      ?= -fPIC
 SLIBCFLAGS      ?=
 CXX             ?= c++ $(ARCHFLAGS)
@@ -84,7 +84,7 @@ FIND_LIBHEADERS := find $(SRCINCDIR) -type f \( -name '*.h' -o \
 LIBHEADERS      := $(shell $(FIND_LIBHEADERS))
 ALLHEADERS      := $(LIBHEADERS) $(wildcard $(SRCDIR)/*.h)
 
-MULTISRCS       := $(SRCDIR)/fdopendir.c
+MULTISRCS       := # Used to have $(SRCDIR)/fdopendir.c because it used struct stat
 ADDSRCS         := $(SRCDIR)/add_symbols.c
 LIBSRCS         := $(filter-out $(MULTISRCS) $(ADDSRCS),$(wildcard $(SRCDIR)/*.c))
 
--- a/README.md
+++ b/README.md
@@ -41,7 +41,10 @@ Wrapped headers and replaced functions are:
   </tr>
   <tr>
     <td><code>dirent.h</code></td>
-    <td>Adds <code>fdopendir</code> function</td>
+    <td>Adds <code>fdopendir</code> function, and wraps <code>opendir</code>,
+    <code>readdir</code>, <code>readdir_r</code>, <code>rewinddir</code>,
+    <code>seekdir</code>, <code>telldir</code>, <code>dirfd</code>, and
+    <code>closedir</code>, to support <code>fdopendir</code></td>
     <td>OSX10.9</td>
   </tr>
   <tr>
--- a/src/best_fchdir.c
+++ b/src/best_fchdir.c
@@ -1,5 +1,6 @@
 /*-
  * Copyright (c) 2019
+ * Copyright (c) 2023 raf <raf@raf.org>
  *
  *
  * THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
@@ -30,7 +31,10 @@ int best_fchdir(int dirfd)
   return syscall(SYS___pthread_fchdir, dirfd);
 #else
 /* Tiger does not have kernel support for __pthread_fchdir, so we have to fall back to fchdir */
-/* unless we can come up with a per-thread compatible implementation that works on Tiger */
+/* unless we can come up with a per-thread compatible implementation that works on Tiger. */
+/* Accept dirfd == -1 (which is meaningful for __pthread_fchdir) but do nothing with it. */
+  if (dirfd == -1)
+    return 0;
   return syscall(SYS_fchdir, dirfd);
 #endif
 }
--- a/test/test_fdopendir.c
+++ b/test/test_fdopendir.c
@@ -1,6 +1,7 @@
 
 /*
  * Copyright (c) 2019
+ * Copyright (c) 2023 raf <raf@raf.org>
  *
  * Permission to use, copy, modify, and distribute this software for any
  * purpose with or without fee is hereby granted, provided that the above
@@ -15,9 +16,11 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
+#include <stdlib.h>
 #include <dirent.h>
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <sys/errno.h>
 #include <unistd.h>
 #include <limits.h>
 #include <fcntl.h>
@@ -55,6 +58,29 @@ int main() {
     }
 
     close(dfd);
+
+    /* Try to use fdopendir with stdin - Should fail with ENOTDIR */
+
+    if ((dir = fdopendir(STDIN_FILENO))) {
+        fprintf(stderr, "fdopendir(stdin) should have failed\n");
+        (void)closedir(dir);
+        return 1;
+    } else if (errno != ENOTDIR) {
+        perror("fdopendir(stdin) should have failed with ENOTDIR");
+        return 1;
+    }
+
+    /* Try to use fdopendir with -1 - Should fail with EBADF */
+
+    if ((dir = fdopendir(-1))) {
+        fprintf(stderr, "fdopendir(-1) should have failed\n");
+        (void)closedir(dir);
+        return 1;
+    } else if (errno != EBADF) {
+        perror("fdopendir(-1) should have failed with EBADF");
+        return 1;
+    }
+
     return 0;
 }
 
--- a/test/test_fmemopen.c
+++ b/test/test_fmemopen.c
@@ -108,10 +108,10 @@ test_autoalloc()
      * Let fmemopen allocate the buffer.
      */
 
-    char str[] = "A quick test";
+    /* char str[] = "A quick test"; */
     FILE *fp;
     long pos;
-    size_t nofw, nofr, i;
+    size_t nofw, i;
     int rc;
 
     /* Open a FILE * using fmemopen. */
--- a/test/test_traverse.c
+++ b/test/test_traverse.c
@@ -1,0 +1,184 @@
+
+/*
+ * Copyright (c) 2023 raf <raf@raf.org>
+ *
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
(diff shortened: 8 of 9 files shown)


kencu:
Thanks for all this!

```__pthread_fchdir()``` is the thread-safe version of ```chdir()```, as you noted, available on 10.5+.

File: src/fdopendir_support.c [hunk 1]
```diff
@@ -0,0 +1,95 @@
...
+}
+
+/*
+ * Wrapped version of dirfd() for fdopendir() compatibility because dirfd()
+ * itself is already a macro (see dirfd macro in dirent.h).
+ */
+
+int __mpls_dirfd(DIR *dirp)
+{
+    return dirfd(dirp->__mpls_dirp);
```
tavianator, line 92: To avoid the confusion of `dirfd(fdopendir(fd)) != fd`, maybe this should do
```c
if (dirp->__mpls_dirfd >= 0) {
    return dirp->__mpls_dirfd;
}
```

kencu:
I am somewhat concerned about dumping pthread_chdir() for chdir()... once again, I don't want to put myself off as some kind of expert, but I read around how these calls are expected to work.

Some of them are supposed to change the working dir for the entire executable, others are specifically set up to not do that, and are supposed to only affect the thread they are running in, and leave the executable's working dir alone...

And I got to wondering if your weird witnessed behaviour of pthread_chdir might have been in some way due to the application using threads, and said threads are not interacting with each other properly, or are racy, or similar thread-related worried.

I mean, after all these years, pthread_chdir() can't really suddenly be found to be non-functional on SnowLeopard. 

In the process of fixing things, we can't make them somehow worse in the process...

include/dirent.h
```diff
@@ -35,6 +44,24 @@ extern DIR *fdopendir(int fd) __DARWIN_ALIAS(fdopendir);
 extern DIR *fdopendir(int fd) __DARWIN_ALIAS_I(fdopendir);
 #endif
 
+extern DIR *__mpls_opendir(const char *name);
+extern int __mpls_closedir(DIR *dirp);
+extern int __mpls_dirfd(DIR *dirp);
+
+#define opendir(name)      __mpls_opendir(name)
```
tavianator, line 51: It is legal for applications to suppress macro expansion when calling `opendir()`, `readdir()` etc., since https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/dirent.h.html says

> The following shall be declared as functions and may also be defined as macros. Function prototypes shall be provided.

Rather than use `(opendir)(".")` for legacysupport to call the underlying implementation, I think it would be better to do this here:

```c
#define opendir __mpls_opendir
#define closedir __mpls_closedir

static inline struct dirent *__mpls_readdir(DIR *dirp) {
    return readdir(dirp->__mpls_dirp);
}
#define readdir __mpls_readdir

...
```

And then in `fdopendir.c` do

```
#undef DIR
#undef opendir
#undef closedir
#undef readdir
...

__MPLS_DIR *__mpls_fdopendir(int dirfd) {
    __MPLS_DIR *ret = malloc(sizeof(*ret);
    __pthread_fchdir(dirfd);
    ret->__mpls_dirp = opendir(".");
    ret->__mpls_dirfd = dirfd;
    __pthread_fchdir(-1);
    return ret;
}
```

src/fdopendir.c
```diff
@@ -49,45 +78,45 @@ DIR *fdopendir(int dirfd) {
...
         return 0;
     }
 
-    dir = opendir (".");
+    dirvp = (void *)(opendir)(".");
 
-    if (best_fchdir(oldCWD) < 0) {
-        if (dir) PROTECT_ERRNO(closedir(dir));
-        if (oldCWD != -1) PROTECT_ERRNO(close(oldCWD));
+    if (fchdir(oldCWD) < 0) {
```
tavianator, line 96: I think this needs to stay as `best_fchdir()`, in order to call `__pthread_fchdir()` if possible.  Otherwise it's no longer thread safe.

tavianator: Actually I think I see why this wasn't working for you, the current (before this patch) `fdopendir()` looks broken to me.  `best_fchdir()` has to be used in [the way that `_ATCALL()` uses it](https://github.com/macports/macports-legacy-support/blob/b570b7f97476e5377ee2397a7a3a71638b72bdb5/src/common-priv.h#L69).  I was going to write a long comment describing how that seems to work, but then I realized you can probably just do this in `fdopendir()`:

```c
DIR *dirp = _ATCALL(dirfd, ".", NULL, opendir("."));
```

That way you don't have to track `oldCWD` yourself or anything.  Should simplify it a lot.

Ionic:
> I am somewhat concerned about dumping pthread_fchdir() for chdir()... once again, I don't want to put myself off as some kind of expert, but I read around how these calls are expected to work.
> 
> Some of them are supposed to change the working dir for the entire executable, others are specifically set up to not do that, and are supposed to only affect the thread they are running in, and leave the executable's working dir alone...

Yes, that's the difference. `fchdir()` itself is totally thread safe, but changes the working directory for all threads of a process.

`__pthread_fchdir()` only changes the working directory for a specific thread.

The `*at` calls were implemented using a thread-local working directory in mind, because, like `fdopendir()`, they internally change the cwd to support relative paths (essentially doing something like `pushd foo; work; popd` in shell parlance). While that could have been implemented with `fchdir()`, and even if `fchdir()` itself is thread safe, we cannot make the internal code atomic and thus thread safe. If `fchdir()` was used, the process's cwd would change and different threads might (briefly) see different cwds as a result. All of that doesn't happen with `__pthread_fchdir()`, which makes the functions actually thread safe - at least in this regard.

kencu:
we're lucky to have a few folks around at the moment who seem to get this -- @tavianator and others, thanks. Perhaps we will get it right this time!

src/fdopendir_support.c
macportsraf: Thanks. It's done. I was wondering which fd would be better to return. Since whatever calls dirfd() isn't allowed to change the state, and both refer to the same directory, I thought it shouldn't really matter which. But if once choice can lead to confusion, then the other choice is better.

include/dirent.h
macportsraf: OK, but I don't understand in what way that is better. It's done.

src/fdopendir.c
macportsraf: Thanks. That's perfect. It fixes the bug and removes code duplication. It's done.

macportsraf:
I've added the list of wrapped functions to the `dirent.h` section in `README.md`.

tavianator (approved):
FWIW this looks good to me.  I don't have a Mac to test on though.

kencu:
I tried first on Tiger, which probably isn't fair as Tiger is different, and got this:
```
/opt/local/bin/gcc-apple-4.2 -isystem tiger_only/include  -c -std=c99 -Iinclude -Os -Wall test/test_traverse.c -o test/test_traverse.o
/opt/local/bin/gcc-apple-4.2 -isystem tiger_only/include  -Llib  test/test_traverse.o -lMacportsLegacySupport -o test/test_traverse
test/test_traverse
fstatat(parent_fd=-2, .test.dir) ok
openat(parent_fd=-2, .test.dir) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) ok
openat(parent_fd=3, subdir) = dir_fd=5 ok
fdopendir(dir_fd=5) ok
entry file
fstatat(parent_fd=5, file) ok
unlink .test.dir/subdir/file failed: No such file or directory
rmdir .test.dir/subdir failed: No such file or directory
rmdir .test.dir failed: No such file or directory
/opt/local/bin/gcc-apple-4.2 -isystem tiger_only/include  -c -std=c99 -Iinclude -Os -Wall test/test_traverse_cwd.c -o test/test_traverse_cwd.o
/opt/local/bin/gcc-apple-4.2 -isystem tiger_only/include  -Llib  test/test_traverse_cwd.o -lMacportsLegacySupport -o test/test_traverse_cwd
test/test_traverse_cwd
mkdir(.test.dir) failed
: File exists
make: *** [Makefile:299: run_traverse_cwd] Error 1
TigerVM:/Users/Shared/test/macports-legacy-support username$ ls
BUILDING.txt    Makefile        include         manual_tests    test
LICENSE         README.md       lib             src             tiger_only
TigerVM:/Users/Shared/test/macports-legacy-support username$ ls -la
total 88
drwxr-xr-x   16 username  wheel    544 Jul 18 21:54 .
drwxr-xr-x    3 username  wheel    102 Jul 18 21:16 ..
drwxr-xr-x   13 username  wheel    442 Jul 18 21:54 .git
-rw-r--r--    1 username  wheel     23 Jul 18 21:16 .gitignore
drwxr-xr-x    3 username  wheel    102 Jul 18 21:54 .test.dir
-rw-r--r--    1 username  wheel     67 Jul 18 21:16 .travis.yml
-rw-r--r--    1 username  wheel   1565 Jul 18 21:16 BUILDING.txt
-rw-r--r--    1 username  wheel   1091 Jul 18 21:16 LICENSE
-rw-r--r--    1 username  wheel  15798 Jul 18 21:29 Makefile
-rw-r--r--    1 username  wheel   8202 Jul 18 21:54 README.md
drwxr-xr-x   28 username  wheel    952 Jul 18 21:54 include
drwxr-xr-x    3 username  wheel    102 Jul 18 21:54 lib
drwxr-xr-x    3 username  wheel    102 Jul 18 21:16 manual_tests
drwxr-xr-x   82 username  wheel   2788 Jul 18 21:54 src
drwxr-xr-x   56 username  wheel   1904 Jul 18 21:54 test
drwxr-xr-x    4 username  wheel    136 Jul 18 21:16 tiger_only
TigerVM:/Users/Shared/test/macports-legacy-support username$ ls .test.dir
subdir
TigerVM:/Users/Shared/test/macports-legacy-support username$ ls .test.dir/subdir
file
TigerVM:/Users/Shared/test/macports-legacy-support username$ 

```

this makes me think on Tiger it is not changing the working directory back to the root directory before running the unlink step....

but, to be honest, breaking Tiger is going to happen sooner or later anyway as it's so different.

Let me try on a couple of newer systems and see what happens with them.

src/fdopendir.c
```diff
@@ -22,72 +23,62 @@
...
+    if (dirfd != AT_FDCWD) {
+        if (fstat(dirfd, &st) < 0)
+            return 0;
 
-    if (dirfd == AT_FDCWD) {
-        dir = opendir (".");
-        /* dirfd can be closed only upon success */
-        if (dir) PROTECT_ERRNO(close(dirfd));
-        return dir;
+        if (!S_ISDIR(st.st_mode)) {
```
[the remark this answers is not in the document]
macportsraf: Thanks! Another brilliant suggestion. I added more tests to `test/test_fdopendir.c` to make sure that the failure modes (e.g., dirfd=0 and dirfd=-1) behave the same with or without that redundant check, and they do. So now the code looks very lean.

Even better, now that `src/fdopendir.c` doesn't use `struct stat` it no longer has to be compiled multiple times for different architectures. It was the only source file that needed that, so the compilation is simpler. I haven't removed from the `Makefile` the framework for compiling multiple times for different architectures (in case a need for it ever arises again), but it currently isn't used.

That means I was able to move the functions in `src/fdopendir_support.c` into `src/fdopendir.c` (so one file fewer).

I also removed two unused variables in `test/test_fmemopen.c` to get rid of warnings.

And I also added `-Wno-deprecated-declarations` to the `Makefile` to suppress warnings about `lstat64` and `stat64` in a few files that included `<sys/stat.h>` (i.e., `src/atcalls.c` and `test/test_fstatat64.c`). If that's not desirable, I can take it out again.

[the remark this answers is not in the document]
kencu: > And I also added `-Wno-deprecated-declarations` to the `Makefile` to suppress warnings about `lstat64` and `stat64` in a few files that included `<sys/stat.h>` (i.e., `src/atcalls.c` and `test/test_fstatat64.c`). If that's not desirable, I can take it out again.

I just have to make sure the compilers we care about (gcc-4.2 and newer, and the system clangs on 10.7+) will swallow that flag.

macportsraf:
@kencu, that's a pity. The test that went wrong first is `test/test_traverse.c` which doesn't chdir at all (at least in the application - the only fchdirs are in `_ATCALL`). The second test (`test/test_traverse_cwd.c`) is the one that chdirs into the `.test.dir` directory first and chdirs back out again at the end. It didn't run at all because the previous test left the `.test.dir` directory behind.

I've just added `system("rm -rf .test.dir")` to the start of both tests so you can see how the second test behaves on 10.4 as well, even after the first test leaves a mess behind.

I've also added some output to the first test to show what the cwd is before and after traversing, but it doesn't (yet) treat a difference as a failure, it just mentions the fact that they are different.

kencu:
I’m tied up for about 48hrs with something so will get back to this then… perhaps another macporter can help run the tests too…

src/fdopendir.c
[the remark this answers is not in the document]
macportsraf: Never mind. It's probably not worth the effort to check. It's just warnings. I've removed it.

kencu:
testing with 10.6.8 Intel 64 bit looks good:

``` 
with the system gcc-4.2:

gcc-4.2 -c -std=c99 -Iinclude -Os -Wall  test/test_traverse.c -o test/test_traverse.o
gcc-4.2 -Llib  test/test_traverse.o -lMacportsLegacySupport -o test/test_traverse
test/test_traverse
cwd (before traverse) /opt/macportsofficial/macports-legacy-support
fstatat(parent_fd=-2, .test.dir) ok
openat(parent_fd=-2, .test.dir) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) ok
openat(parent_fd=3, subdir) = dir_fd=5 ok
fdopendir(dir_fd=5) ok
entry file
fstatat(parent_fd=5, file) ok
cwd (after traverse)  /opt/macportsofficial/macports-legacy-support
gcc-4.2 -c -std=c99 -Iinclude -Os -Wall  test/test_traverse_cwd.c -o test/test_traverse_cwd.o
gcc-4.2 -Llib  test/test_traverse_cwd.o -lMacportsLegacySupport -o test/test_traverse_cwd
test/test_traverse_cwd
cd .test.dir
fstatat(parent_fd=-2, .) ok
openat(parent_fd=-2, .) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) ok
openat(parent_fd=3, subdir) = dir_fd=5 ok
fdopendir(dir_fd=5) ok
entry file
fstatat(parent_fd=5, file) ok
cwd (before cd ..) /opt/macportsofficial/macports-legacy-support/.test.dir
cwd (after cd ..)  /opt/macportsofficial/macports-legacy-support
gcc-4.2 -c -std=c99 -Iinclude -Os -Wall  test/test_usb.c -o test/test_usb.o
gcc-4.2 -Llib  test/test_usb.o -lMacportsLegacySupport -o test/test_usb


with the current clang-mp-11:

clang-mp-11 -c -std=c99 -Iinclude -Os -Wall  test/test_traverse.c -o test/test_traverse.o
In file included from test/test_traverse.c:53:
include/dirent.h:56:13: warning: extra tokens at end of #undef directive [-Wextra-tokens]
#undef dirfd(dirp)
            ^
            //
1 warning generated.
clang-mp-11 -Llib  test/test_traverse.o -lMacportsLegacySupport -o test/test_traverse
test/test_traverse
cwd (before traverse) /opt/macportsofficial/macports-legacy-support
fstatat(parent_fd=-2, .test.dir) ok
openat(parent_fd=-2, .test.dir) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) ok
openat(parent_fd=3, subdir) = dir_fd=5 ok
fdopendir(dir_fd=5) ok
entry file
fstatat(parent_fd=5, file) ok
cwd (after traverse)  /opt/macportsofficial/macports-legacy-support
clang-mp-11 -c -std=c99 -Iinclude -Os -Wall  test/test_traverse_cwd.c -o test/test_traverse_cwd.o
In file included from test/test_traverse_cwd.c:53:
include/dirent.h:56:13: warning: extra tokens at end of #undef directive [-Wextra-tokens]
#undef dirfd(dirp)
            ^
            //
1 warning generated.
clang-mp-11 -Llib  test/test_traverse_cwd.o -lMacportsLegacySupport -o test/test_traverse_cwd
test/test_traverse_cwd
cd .test.dir
fstatat(parent_fd=-2, .) ok
openat(parent_fd=-2, .) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) ok
openat(parent_fd=3, subdir) = dir_fd=5 ok
fdopendir(dir_fd=5) ok
entry file
fstatat(parent_fd=5, file) ok
cwd (before cd ..) /opt/macportsofficial/macports-legacy-support/.test.dir
cwd (after cd ..)  /opt/macportsofficial/macports-legacy-support
```

I suppose we might silence that #undef error in case some build has -Werror enabled.

kencu:
and testing on 10.6.8 64 bit building as 32bit also looks good:

```
gcc-4.2 -arch i386  -c -std=c99 -Iinclude -Os -Wall  test/test_traverse.c -o test/test_traverse.o
gcc-4.2 -arch i386  -Llib  test/test_traverse.o -lMacportsLegacySupport -o test/test_traverse
test/test_traverse
cwd (before traverse) /opt/macportsofficial/macports-legacy-support
fstatat(parent_fd=-2, .test.dir) ok
openat(parent_fd=-2, .test.dir) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) ok
openat(parent_fd=3, subdir) = dir_fd=5 ok
fdopendir(dir_fd=5) ok
entry file
fstatat(parent_fd=5, file) ok
cwd (after traverse)  /opt/macportsofficial/macports-legacy-support
gcc-4.2 -arch i386  -c -std=c99 -Iinclude -Os -Wall  test/test_traverse_cwd.c -o test/test_traverse_cwd.o
gcc-4.2 -arch i386  -Llib  test/test_traverse_cwd.o -lMacportsLegacySupport -o test/test_traverse_cwd
test/test_traverse_cwd
cd .test.dir
fstatat(parent_fd=-2, .) ok
openat(parent_fd=-2, .) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) ok
openat(parent_fd=3, subdir) = dir_fd=5 ok
fdopendir(dir_fd=5) ok
entry file
fstatat(parent_fd=5, file) ok
cwd (before cd ..) /opt/macportsofficial/macports-legacy-support/.test.dir
cwd (after cd ..)  /opt/macportsofficial/macports-legacy-support
```

kencu:
10.5 is good for archs i386, x86_64, and ppc.

when trying to run the test suite using ppc64, however, I get this error, which has nothing to do with this PR:

```
gcc-4.2 -arch ppc64  -dynamiclib -headerpad_max_install_names -install_name @executable_path/../lib/libMacportsLegacySupport.dylib -current_version 1.0 -compatibility_version 1.0  src/arc4random.dl.o src/atcalls.dl.o src/best_fchdir.dl.o src/dprintf.dl.o src/fdopendir.dl.o src/fmemopen.dl.o src/fsetattrlist.dl.o src/fsgetpath.dl.o src/getdelim.dl.o src/getentropy.dl.o src/getline.dl.o src/lchmod.dl.o src/lutimes.dl.o src/macports_legacy_atexit.dl.o src/macports_legacy_copyfile.dl.o src/macports_legacy_os_unfair_lock.dl.o src/macports_legacy_realpath.dl.o src/macports_legacy_sysconf.dl.o src/memmem.dl.o src/memstream.dl.o src/posix_memalign_emulation.dl.o src/pthread_setname_np.dl.o src/setattrlistat.dl.o src/sincos.dl.o src/strings.dl.o src/strndup.dl.o src/strnlen.dl.o src/time.dl.o src/utimensat.dl.o src/wcpcpy_family.dl.o src/wcscasecmp_family.dl.o src/wcsdup.dl.o src/wcsnlen.dl.o -o lib/libMacportsLegacySupport.dylib
gcc-4.2 -arch ppc64  -Llib  test/test_arc4random.o -lMacportsLegacySupport -o test/test_arc4random
test/test_arc4random
make: execvp: test/test_arc4random: Bad CPU type in executable
make: *** [run_arc4random] Error 127
LeopardIntelServerVM:macports-legacy-support username$ file test/test_arc4random
test/test_arc4random: Mach-O 64-bit executable ppc64
LeopardIntelServerVM:macports-legacy-support username$ test/test_arc4random
-bash: test/test_arc4random: Bad CPU type in executable
```

this is a bit odd, as I thought Leopard Intel was supposed to run ppc64, but I haven't really tested it much at all. That will have to be tested on the PPC machine.

kencu:
for anyone who would like to run some tests themselves, this is the basic format:
```
  502  cd /tmp
  503  git clone https://github.com/macports/macports-legacy-support
  504  cd mac*
  505  git fetch origin pull/61/head:pull61
  506  git branch
  507  git checkout pull61
  508  CC=gcc-4.2 CXX=g++-4.2 make test
```

Tiger needs the extra bit, as above.

kencu:
10.7 is good with the system clang both i386 and x86_64

include/dirent.h
```diff
@@ -35,7 +44,47 @@ extern DIR *fdopendir(int fd) __DARWIN_ALIAS(fdopendir);
...
+/* Wrapper functions/macros to support fdopendir */
+extern DIR *__mpls_opendir(const char *name);
+extern int __mpls_closedir(DIR *dirp);
+extern int __mpls_dirfd(DIR *dirp);
+
+#define opendir(name)      __mpls_opendir(name)
+#define closedir(dirp)     __mpls_closedir(dirp)
+
+#ifndef __MP_LEGACY_SUPPORT_NO_DIRFD_MACRO
+#undef dirfd(dirp)
```
kencu, line 56: This should be just
```
#undef dirfd
```
otherwise warnings

kencu:
10.8 is good.

kencu:
10.4 (i386 tested) is still broken though.
```
/opt/local/bin/gcc-apple-4.2 -isystem tiger_only/include  -c -std=c99 -Iinclude -Os -Wall  test/test_traverse.c -o test/test_traverse.o
/opt/local/bin/gcc-apple-4.2 -isystem tiger_only/include  -Llib  test/test_traverse.o -lMacportsLegacySupport -o test/test_traverse
test/test_traverse
cwd (before traverse) /private/tmp/macports-legacy-support
fstatat(parent_fd=-2, .test.dir) ok
openat(parent_fd=-2, .test.dir) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) ok
openat(parent_fd=3, subdir) = dir_fd=5 ok
fdopendir(dir_fd=5) ok
entry file
fstatat(parent_fd=5, file) ok
cwd (after traverse)  /private/tmp/macports-legacy-support/.test.dir/subdir
Directory has changed while traversing!
unlink .test.dir/subdir/file failed: No such file or directory
rmdir .test.dir/subdir failed: No such file or directory
rmdir .test.dir failed: No such file or directory
/opt/local/bin/gcc-apple-4.2 -isystem tiger_only/include  -c -std=c99 -Iinclude -Os -Wall  test/test_traverse_cwd.c -o test/test_traverse_cwd.o
/opt/local/bin/gcc-apple-4.2 -isystem tiger_only/include  -Llib  test/test_traverse_cwd.o -lMacportsLegacySupport -o test/test_traverse_cwd
test/test_traverse_cwd
cd .test.dir
fstatat(parent_fd=-2, .) ok
openat(parent_fd=-2, .) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) ok
openat(parent_fd=3, subdir) = dir_fd=5 ok
fdopendir(dir_fd=5) ok
entry file
fstatat(parent_fd=5, file) ok
cwd (before cd ..) /private/tmp/macports-legacy-support/.test.dir/subdir
cwd (after cd ..)  /private/tmp/macports-legacy-support/.test.dir
unlink .test.dir/subdir/file failed: No such file or directory
rmdir .test.dir/subdir failed: No such file or directory
rmdir .test.dir failed: No such file or directory
```

@evanmiller -- any chance you might look at this and see what those sharp eyes of yours can see?

kencu:
It looks like even the oldest compilers accept "-Wno-deprecated-declarations".

kencu:
So this is going to be a big problem for Tiger, it seems... here's some background.

https://stackoverflow.com/questions/13438910/if-i-chdir-within-a-thread-will-that-affect-the-cwd-of-the-parent-program?rq=3

include/dirent.h
macportsraf: Thanks. It's done.

macportsraf:
> It looks like even the oldest compilers accept "-Wno-deprecated-declarations".

Not necessarily. I commented that out just in case it might be a problem.

macportsraf:
On 10.4, it looks like the fchdir in _ATCALL that go into each subdirectory works, but the fchdirs that come back out again don't. Actually, looking at _ATCALL, that makes sense. Before the first best_fchdir(dirfd), there is another wierd call that I don't understand the point of:

            oldCWD = open(".", O_RDONLY);                       \
            if (best_fchdir(-1) < 0 && oldCWD != -1) {          \
                close(oldCWD); oldCWD = -1;                     \
            }                                                   \

I can't imagine a situation where fchdir(-1) wouldn't return -1 (with errno == EBADF), so on 10.4, it loses the knowledge of where to fchdir back to. Presumably __pthread_fchdir (on 10.5+) doesn't return -1 when dirfd == -1.

Does anyone know why this if statement is there? It's not documented in the code itself. I wouldn't be suprised if it would start working if that if statement were removed on 10.4, but it would be nice to know why it's there.

Should we give that a go (just on 10.4, that is)?

Ionic:
>I can't imagine a situation where fchdir(-1) wouldn't return -1 (with errno == EBADF), so on 10.4, it loses the knowledge of where to fchdir back to. Presumably __pthread_fchdir (on 10.5+) doesn't return -1 when dirfd == -1 (no doubt for some good reason that I'm unaware of).

Oh, yes, absolutely true. `__pthread_fchdir (-1)` resets the per-thread directory for the current thread and uses the process-global cwd again.

`fchdir ()` will just return an error.

So, for this usage, they aren't drop-in replacements.

Bummer, it really looks like the macro has to be copied for 10.4 and the call removed, like you already did.

The intention is probably to reset the thread's CWD to whatever the process uses, in case it was overridden on a per-thread basis. We can't know what other code might have done, so the precaution generally seems to make sense. To be sure that `*at` calls are simulated correctly, we'd have to check what Apple's implementation does if the per-thread CWD doesn't match the process CWD and such an `*at` call is issued.

If Apple's implementation doesn't reset the per-thread CWD, legacy-support probably shouldn't do it either. If it does, the current code is fine (at least for anything non-10.4).

kencu:
> > It looks like even the oldest compilers accept "-Wno-deprecated-declarations".
> 
> Not necessarily. I commented that out just in case it might be a problem.

I added it again.

kencu:
That latest change appears to fix Tiger (i386 tested)... woo-hoo

```
/opt/local/bin/gcc-apple-4.2 -isystem tiger_only/include  -c -std=c99 -Iinclude -Os -Wall  test/test_traverse.c -o test/test_traverse.o
/opt/local/bin/gcc-apple-4.2 -isystem tiger_only/include  -Llib  test/test_traverse.o -lMacportsLegacySupport -o test/test_traverse
test/test_traverse
cwd (before traverse) /Users/Shared/test/macports-legacy-support
fstatat(parent_fd=-2, .test.dir) ok
openat(parent_fd=-2, .test.dir) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) ok
openat(parent_fd=3, subdir) = dir_fd=6 ok
fdopendir(dir_fd=6) ok
entry file
fstatat(parent_fd=6, file) ok
cwd (after traverse)  /Users/Shared/test/macports-legacy-support
/opt/local/bin/gcc-apple-4.2 -isystem tiger_only/include  -c -std=c99 -Iinclude -Os -Wall  test/test_traverse_cwd.c -o test/test_traverse_cwd.o
/opt/local/bin/gcc-apple-4.2 -isystem tiger_only/include  -Llib  test/test_traverse_cwd.o -lMacportsLegacySupport -o test/test_traverse_cwd
test/test_traverse_cwd
cd .test.dir
fstatat(parent_fd=-2, .) ok
openat(parent_fd=-2, .) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) ok
openat(parent_fd=3, subdir) = dir_fd=6 ok
fdopendir(dir_fd=6) ok
entry file
fstatat(parent_fd=6, file) ok
cwd (before cd ..) /Users/Shared/test/macports-legacy-support/.test.dir
cwd (after cd ..)  /Users/Shared/test/macports-legacy-support
```

kencu:
I suppose if we are going to have two ATCALL macros, one for 10.4 and one for 10.5+, we could then just get rid of the best_fchdir confusion, and use the actual calls in the macros.

jmroot:
> I thought Leopard Intel was supposed to run ppc64

No, Rosetta only runs ppc32 binaries.

macportsraf:
Rather than having to versions of _ATCALL, I think it's cleaner to just make best_fchdir(-1) return 0 on 10.4. Then it's only a 2-line change.

I've also put `-Wno-deprecated-declarations` back into the Makefile.

Does it still work?

kencu:
I think you should be able to test the Tiger pathway on any system by adjusting the best_fchdir macro temporarily… should make this easier

kencu:
someday we might have a test that does some work in a pthread… but for now, ignorance is bliss.

all done, ready to commit?

mascguy:
Awesome work and collaboration on this!

macportsraf:
> I think you should be able to test the Tiger pathway on any system by adjusting the
> best_fchdir macro temporarily… should make this easier

Thanks. Yes. That's how I knew that the fdopendir(-1) test would fail. I temporarily made the 10.5+ path return 0 when dirfd == -1.

I've tidied up the code a bit, and I think it's ready, but it would be great if it can be tested again at least on 10.4 and maybe 10.5. I've tested it on 10.6.

Many thanks to everyone for all the help. I'm so glad that I'll be able to use rawhide on my 10.6 laptop. Macports is great!

kencu:
10.4 and 10.5 all pass.

mascguy:
Thanks again for all of your great work on this Raf, as well as the help and support of Ken and others!

PR #63, opened 2023-08-18
fstatat64: tests: add declaration for fstatat64 syscall, to fix build issues for newer macOS releases

mascguy:
On newer macOS releases - like, say, 10.15 - the test for `fstatat64` fails to build:

```
test/test_fstatat64.c:93:12:
error: implicit declaration of function 'fstatat64' is invalid in C99

if (0 != fstatat64 (AT_FDCWD, argv[0], &atfdcwd_stat64, 0)) {
         ^
```

Since this syscall is provided via header `sys/sysproto.h` - within `Kernel.framework` - it's not ideal to depend on the latter. So rather than going down that path, simply add a local declaration for `fstatat64`.

Fixes: [Ticket 65913 - legacy-support: tests: failure due to fstatat64](https://trac.macports.org/ticket/65913)

### DIFF
--- a/test/test_fstatat64.c
+++ b/test/test_fstatat64.c
@@ -26,6 +26,14 @@
 
 #define FEEDBACK 1
 
+/*
+ * For newer macOS releases - where this syscall is available - it's declared
+ * in header 'sys/sysproto.h', within 'Kernel.framework'.
+ *
+ * To avoid depending on that, declare it ourselves. (Remember that C/C++
+ * allows redundant function declarations, as long as the signatures match.)
+ */
+int fstatat64(int dirfd, const char *pathname, struct stat64 *buf, int flags);
 
 int main (int argc, char **argv) {
 #if !__DARWIN_ONLY_64_BIT_INO_T && __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ >= 1050


cjones051073:
If the method is provided by the system, then surely there must be header that defines it ?

I don’t particularly like the solution as here..

Ionic:
I concur. The proper header inclusion is probably just missing, and that can easily be made dependent on the OS version.

mascguy:
The function is declared in a header that specifically lives in `Kernel.framework`.

Are we sure we want to deal with adding a selective dependency on that, depending on the version of macOS?

That wouldn't be my choice.

Regardless, standard C/C++ allows duplicate/redundant function declarations, as long as they don't conflict with one another. Which is why I chose this path.

I'm open to other ideas though...

mascguy:
Alternatively, we could simply disable the test for this function, for macOS releases where it's not being provided by `legacy-support`.

Though that would prevent validation of a given test's behavior, against the macOS-provided version.

cjones051073:
> Alternatively, we could simply disable the test for this function, for macOS releases where it's not being provided by `legacy-support`.
> 
> Though that would prevent validation of a given test's behavior, against the macOS-provided version.
> 
> That was another option I ruled out, given the consequences.

Yes, this is not ideal as comparing the tests on platforms where legacy-support does, and does not provide the function is very useful.

mascguy:
FYI, I'll add a comment regarding the reason for the declaration, to help answer these questions too. (That would have helped originally. Lol)

mascguy:
Also, this where I'm seeing the header (for 10.15, with Xcode 12.0):

```
$ gfind /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer -type f -name "*.h" -print0 | gxargs -0 -r ggrep -E  "fstatat64\("

/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Kernel.framework/Versions/A/Headers/sys/sysproto.h:int fstatat64(struct proc *, struct fstatat64_args *, int *);
```

mascguy:
> FYI, I'll add a comment regarding the reason for the declaration, to help answer these questions too. (That would have helped originally. Lol)

Done.

mascguy:
Is this slightly more palatable, with explanation comment?

Whatever we do, we need to ensure that our tests run across-the-board, rather than fail with a build error on some releases. So let's make a decision, and get this fixed please...

cjones051073:
Just curious, but what happens if you try and just include the header sys/sysproto.h ?

mascguy:
> Just curious, but what happens if you try and just include the header sys/sysproto.h ?

Adding an include for that header - without touching the compilation flags - results in failure.

Testing on both 10.9 and 10.12, we end up with:

```
$ sudo port -v test legacy-support-devel

[...]

test/test_fstatat64.c:17:10: fatal error: 'sys/sysproto.h' file not found

#include <sys/sysproto.h>
         ^
```

The header is indeed present in `Kernel.framework` for both. But without adding compile option `-framework Kernel` - and updating appropriate env vars, etc, to ensure the macOS SDK is found - those headers aren't in the search path.

cjones051073:
OK, good enough for now I guess.

PR #65, opened 2023-09-01
Ignore unimplemented AT_EACCESS flag, trac bugs #65569 and #67406

acjones8:
Emacs currently doesn't compile on older versions of OS X because legacy-support doesn't implement the `AT_EACCESS` flag in `faccessat()` and instead returns an invalid argument error, which causes Emacs to crash during the phase where it tries to dump to disk. This patch instead zeros out the flag if it detects that it's `AT_EACCESS`, and will then attempt to access the file with the real user's permissions instead of the effective user. This allows Emacs to compile (since the macports user has access to its own build directory), and in my testing, I have yet to encounter any runtime issues. This should also fix any other program that has permission to access a file or directory with its real user's permissions, but fails just because it attempts to use `AT_EACCESS()` to get the effective user's permissions.

### DIFF
--- a/src/atcalls.c
+++ b/src/atcalls.c
@@ -104,6 +104,10 @@ int faccessat(int dirfd, const char *pathname, int mode, int flags)
     } __attribute__((aligned(4), packed)) ab;
     unsigned long opts = 0;
 
+    // Ignore AT_EACCESS, Bug #65569 and #67406
+    if (flags == AT_EACCESS)
+        flags = 0; // Zero out flag, ignore request to use effective user ID
+
     ERR_ON(EINVAL, flags & ~AT_SYMLINK_NOFOLLOW);
     if (flags & AT_SYMLINK_NOFOLLOW)
         opts |= FSOPT_NOFOLLOW;


barracuda156:
@macportsraf @mascguy @catap Could someone review this please?

catap:
@acjones8 you simple broke feature. `faccessat(AT_EACCESS)` should work near the same way with `euidaccess`. I've make a quick github search and discovers this pice of code https://github.com/att/ast/blob/master/src/lib/libast/comp/eaccess.c#L43-L137 which implements logic which you're ignoring.

catap:
After reading sources of maCOS 10.10: https://github.com/apple-open-source/macos/blob/10.10/xnu/bsd/vfs/vfs_syscalls.c#L5130-L5219 I'd like to suggest patch:

```diff
diff --git a/src/atcalls.c b/src/atcalls.c
index 8b9bbe8..a26140b 100644
--- a/src/atcalls.c
+++ b/src/atcalls.c
@@ -104,12 +104,81 @@ int faccessat(int dirfd, const char *pathname, int mode, int flags)
     } __attribute__((aligned(4), packed)) ab;
     unsigned long opts = 0;
 
+    bool has_at_eaccess;
+
+    if (flags & AT_EACCESS)
+    {
+        // AT_EACCESS aren't supported anywhere else, remove it
+        flags &= ~AT_EACCESS;
+        has_at_eaccess = true;
+    }
+
     ERR_ON(EINVAL, flags & ~AT_SYMLINK_NOFOLLOW);
     if (flags & AT_SYMLINK_NOFOLLOW)
         opts |= FSOPT_NOFOLLOW;
     if (getattrlistat(dirfd, pathname, &al, &ab, sizeof(ab), opts) < 0)
         return -1;
     ERR_ON(EACCES, (ab.access & mode) != (uint32_t)mode);
+    if (has_at_eaccess)
+    {
+        int	mode;
+        uid_t   ruid = getuid();
+        uid_t   euid = geteuid();
+        gid_t   rgid = getgid();
+        gid_t   egid = getegid();
+        if (ruid == euid && rgid == egid)
+            return 0;
+
+        struct stat st;
+        if (stat(path, &st) < 0)
+            return -1;
+
+        if (euid == 0)
+            ERR_ON(EACCES, S_ISREG(st.st_mode) && (flags & X_OK) && !(st.st_mode & (S_IXUSR|S_IXGRP|S_IXOTH)));
+        else if (euid == st.st_uid)
+        {
+            if (flags & R_OK)
+                mode |= S_IRUSR;
+            if (flags & W_OK)
+                mode |= S_IWUSR;
+            if (flags & X_OK)
+                mode |= S_IXUSR;
+        }
+        else if (egid == st.st_gid)
+        {
+            if (flags & R_OK)
+                mode |= S_IRGRP;
+            if (flags & W_OK)
+                mode |= S_IWGRP;
+            if (flags & X_OK)
+                mode |= S_IXGRP;
+        }
+        else
+        {
+            int     n;
+            gid_t   groups[NGROUPS_MAX];
+
+            n = getgroups(NGROUPS_MAX, groups);
+            while (--n >= 0)
+                if (groups[n] == st.st_gid)
+                {
+                    if (flags & R_OK)
+                        mode |= S_IRGRP;
+                    if (flags & W_OK)
+                        mode |= S_IWGRP;
+                    if (flags & X_OK)
+                        mode |= S_IXGRP;
+                }
+
+            if (flags & R_OK)
+                mode |= S_IROTH;
+            if (flags & W_OK)
+                mode |= S_IWOTH;
+            if (flags & X_OK)
+                mode |= S_IXOTH;
+        }
+        ERR_ON(EACCES, (st.st_mode & mode) != mode)
+    }
     return 0;
 }
```

I haven't tried to compile it, just make it. Probably it needs some polish.

Feel free to adopt it or use it, or dismiss it.

Here I've tried my best to implement expected behaviour of `faccessat(AT_EACCESS)`.

acjones8:
Thanks for the patch, yours (which actually implements the flag instead of just working around it) is much better. I don't have very much experience with systems programming like this, so I wasn't sure how to implement it.

I gave it a try, and I only needed to fix two things to get it to compile. The first is to surround the `ERR_ON` right underneath `if (euid == 0)`  with braces, so it looks like:
```
if (euid == 0)
{
    ERR_ON(EACCES, S_ISREG(st.st_mode) && (flags & X_OK) && !(st.st_mode & (S_IXUSR|S_IXGRP|S_IXOTH)));
}
```
The second fix I had to do was to change `if (stat(path, &st) < 0)` to `if (stat(pathname, &st) < 0)`. After that, the code compiles just fine. I rebuilt Emacs with the new version of legacy-support, and it also compiles and functions perfectly. It would be great to see this patch added, so that Emacs works out of the box again on older systems.

barracuda156:
@acjones8 I guess you could rebase your PR to Kirill’s patch, referencing him.

It is desirable to test this prior to adding to `legacy-support`, of course, since it may have effects not only for EMacs.

@mascguy @ryandesign @jmroot @macportsraf Any thoughts on this?

macportsraf:
Can it affect any ports that don't refer to `AT_EACCESS`? Surely they wouldn't have compiled.

It's definitely a good idea to add a unit test that fails without this change and passes with it.

catap:
@macportsraf well.. `faccessat ` was introduced at 10l.10 and had a code inside:
```c
int
faccessat(__unused proc_t p, struct faccessat_args *uap,
          __unused int32_t *retval)
{
	if (uap->flag & ~AT_EACCESS)
		return (EINVAL);

	return (faccessat_internal(vfs_context_current(), uap->fd,
	    uap->path, uap->amode, uap->flag, UIO_USERSPACE));
}
```
So, literally it means that `AT_EACCESS` is mandatory flag.

Inside `faccessat_internal` it uses only to decide which `context.vc_ucred` is used and my pice of code tries to emulate logic of used context when such flag is provided as good as I can.

So, it seems that current implementation isn't right and it should behaves differently on 10.10 and 10.9, but I haven't tested it.

@acjones8 if you ok, I may bring everything together as dedicated branch (which you need to consume in this PR :) ) which will include trivial test.

Trivial means that it should confirm that it behaviour near the same, but logic is quite complicated, requires some additional groups and so on => and test it is quite a challenge.

macportsraf:
I've written tests (attached) for `faccessat`, but I'm having trouble with something. To test it properly, it needs to be tested as a non-setuid program (the default), and as a setuid program (which `make test` will arrange if run by root).

The problem is that the test program can't load the *LegacySupport* library being tested because:

    dyld: Library not loaded: @executable_path/../lib/libMacportsLegacySupport.dylib
    Referenced from: /Users/raf/src/macports-legacy-support/test/test_faccessat_setuid
    Reason: unsafe use of @executable_path in /Users/raf/src/macports-legacy-support/test/test_faccessat_setuid with restricted binary

The linking needs to be against a library with an absolute path. The library is allowed to load when I link against the currently installed *LegacySupport* library:

    cc -L/opt/local/lib  test/test_faccessat.o -lMacportsLegacySupport -o test/test_faccessat
    cp test/test_faccessat test/test_faccessat_setuid
    sudo chown root test/test_faccessat_setuid
    sudo chmod 4755 test/test_faccessat_setuid 
    test/test_faccessat_setuid
    faccessat_setuid: uid=501 euid=0

So I thought that the problem would be fixed by changing the `Makefile` from:

    TESTLDFLAGS = -L$(BUILDDLIBDIR) $(LDFLAGS)

to:

    TESTLDFLAGS = -L"$(shell pwd)"/$(BUILDDLIBDIR) $(LDFLAGS)

But that didn't fix it. `otool -L test/test_faccessat_setuid` still says:

    @executable_path/../lib/libMacportsLegacySupport.dylib (compatibility version 1.0.0, current version 1.0.0)
    /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.0.0)

The linking command was:

    cc -L"/Users/raf/src/macports-legacy-support"/lib  test/test_faccessat.o -lMacportsLegacySupport -o test/test_faccessat

Any idea why that didn't work, and what I can do to let the setuid test load the library being tested?

Is it because `/opt/local/lib` is only writable by root but `/Users/...` isn't? No. I tried building everything as root in a rootonly-writable directory hierarchy and it didn't change the *otool* output.

P.S. I was wrong earlier about ports that use `AT_EACCESS` not compiling. They compile but might not work. When run as a non-root user on 10.6, this test shows that a readable file is reported as unreadable with `AT_EACCESS` (uid==euid==fileowner).

P.P.S. Even without fixing the problem with the setuid test, these tests already demonstrate that `AT_EACCESS` doesn't work on 10.6, so I think I should submit it as a pull request. But I'll wait to see if anyone can solve the problem with the setuid test.

P.P.P.S These tests work on 10.14, but not on 10.6. Here are those test results:

Not setuid and run by non-root user:

    not root: faccessat(readable, AT_EACCESS) is unreadable = FAIL
    -r--------  1 raf  staff  0  3 Sep 19:26 test/tmp.readable
    ----------  1 raf  staff  0  3 Sep 19:26 test/tmp.unreadable

Not setuid and run by root:

    root: faccessat(readable, AT_EACCESS) is unreadable = FAIL
    root: faccessat(unreadable, AT_EACCESS) is unreadable = FAIL
    -r--------  1 root  staff  0  3 Sep 19:27 test/tmp.readable
    ----------  1 root  staff  0  3 Sep 19:27 test/tmp.unreadable

Setuid (linked against installed *LegacySupport* library) and run as *nobody*:

    faccessat_setuid: uid=-2 euid=0
    setuid root: not root: faccessat(readable) is readable = FAIL
    setuid root: not root: faccessat(unreadable) is readable = FAIL
    setuid root: not root: faccessat(readable AT_EACCESS) is unreadable = FAIL
    setuid root: not root: faccessat(unreadable AT_EACCESS) is unreadable = FAIL
    -r--------  1 root  staff  0  3 Sep 19:28 test/tmp.readable
    ----------  1 root  staff  0  3 Sep 19:28 test/tmp.unreadable

I'm surprised that the two setuid tests without `AT_EACCESS` failed. Maybe there's something wrong with the test, but I can't see it.

[Makefile.test_faccessat_setuid.patch](https://github.com/macports/macports-legacy-support/files/12505650/Makefile.test_faccessat_setuid.patch)
[test_faccessat.c.txt](https://github.com/macports/macports-legacy-support/files/12505652/test_faccessat.c.txt)

catap:
> Any idea why that didn't work, and what I can do to let the setuid test load the library being tested?

Just a guess: I feel that setuid program should load only libraries by absolute path and all of them should have setuid flag.

Why? To prevent injection / replacement of such library by non privileged user and inject some bad code.

catap:
P.S. if @acjones8 or @macportsraf willing to combinate everything into one nice PR feel free to treat my pice of code under public domain.

macportsraf:
@catap

    Just a guess: I feel that setuid program should load only libraries by absolute path and all of them should have setuid flag.

    Why? To prevent injection / replacement of such library by non privileged user and inject some bad code.

I'm sure that setuid programs can only load libraries via absolute paths, and for the reason you state, but libraries can't be setuid. Only the programs themselves can be.

I came to the same conclusion which is why I tried to link against the library-being-tested using an absolute path, but that attempt didn't work. That's what I'm asking for a solution for. I used `-L"$(shell pwd)"/lib` rather than `-Llib`, but the resulting `otool -L` output still said that it would load the library via `@executable_path` rather than by the absolute path that I had provided. But it depends on which path. If I specify the absolute path `/opt/local/lib`, that works (but doesn't test the library-being-tested). If I specify the absolute path to the `lib` directory in the current directory, it doesn't work.

If nobody can see how to fix it, the tests are still useful without setuid, but not as useful. I could even get it to link against the `/opt/local/lib` version of the library (when possible) to do setuid tests but that won't test the library-being-tested. But I'm hoping someone will see what I've done wrong.

I've improved the tests so that as well as the setuid root test, there's now also a setuid non-root user test. They are different cases.

I can start a pull request that combines the tests and the code above, but it won't address the non-`AT_EACCESS` tests that failed. More research will be needed for that.

catap:
I think that to start from test which covers `faccessat` and to confirm that it behaves the same way on 10.10+ and on 10.6 is very good step.

And I hope that it can be merged quite fast.

macportsraf:
if (ruid == euid && rgid == egid)
        return 0;

I don't think the above is right. If the program is not setuid or setgid, this returns success before looking at the file. I think it should be moved up to where `AT_EACCESS` is detected and removed, and if this condition is true, just forget that `AT_ACCESS` was there at all. The result for the effective ids will be the same as the result for the real ids.

I'll keep working on it tomorrow.

catap:
@macportsraf as I state before, I just write the code. I haven't even tried to compile it. Yes, you defently right with this assumption, it is an issue in my original logic.

Also, keep in mind that you have not only euid/egid, you also have multiple groups on the user.

macportsraf:
Yes, my tests don't cover supplementary groups. I don't think I can test for them without manipulating users and groups to suit the needs of the test. I'm not sure I want to try that on macOS. However, I don't think there are any supplementary groups when a program is setuid or setgid. But I'll check to make sure. Perhaps that code won't be needed.

macportsraf:
Update: I've submitted a new pull request (https://github.com/macports/macports-legacy-support/pull/66) with a new implementation of `faccessat`. The existing implementation didn't work at all in setuid/setgid programs where it matters. It also didn't support `AT_EACCESS` or `F_OK`. The new implementation is better, but I'm still having trouble linking against the library via an absolute path for setuid testing.

mascguy:
So is this still a work-in-progress...?

catap:
I this this one should be closed

macportsraf:
I agree that it can be closed. The work in progress is #66

mascguy:
> I agree that it can be closed. The work in progress is #66

Great, thanks for the confirmation!

PR #62, opened 2023-07-26
Fix dirent.h for C++ (argless seekdir macro clashed with typename)

macportsraf:
This pull request is an attempt to address https://github.com/vectorgraphics/asymptote/issues/386 which was caused by the recent `fdopendir` fix. The new argumentless `seekdir` macro was clashing with the `seekdir` component of the `ios_base::seekdir` typename.

This pull request changes the argumentless macro definition in `<dirent.h>` so that they include the arguments. But it only does this for C++. The macros remain unchanged (i.e., argumentless) for C. This means that `seekdir` without arguments will only invoke the macro expansion in C, not in C++. The macro expansion will only be invoked in C++ when the parameters are included.

This means that C programs can take the address of functions like seekdir, but C++ programs can't. This might cause problems for some C++ programs but hopefully not. However, if C++ declares a type for seekdir-like functions, perhaps it is necessary to be able to take the address of the seekdir function. Maybe this is wishful thinking.

PR #66, opened 2023-09-04
Fix faccessat (setuid, setgid, AT_EACCESS, F_OK)

macportsraf:
This pull request comes from https://github.com/macports/macports-legacy-support/pull/65.

@acjones8 realised that `faccessat` didn't work with `AT_EACCESS`. It just returns -1 with errno set to `EINVAL`. That prevents emacs from installing on old systems. They suggested silently ignoring `AT_EACCESS`.

@catap contributed an approximate implementation to actually implement `AT_EACCESS` instead.

I wrote some tests for `faccessat` and fixed the new implementation.

I realised that the existing `faccessat` implementation didn't actually work at all in setuid/setgid programs where it really matters. It was always checking effective user access, not real user access, which is what `faccessat` is supposed to check by default. So it only works in non-setuid non-setgid programs where the real user and the effective user are the same. This limitation could lead to security flaws in setuid/setgid programs that use `faccessat`.

This pull request replaces the existing incorrect `faccessat` implementation with a new one that mostly works. It includes decent tests, but not super aggressive tests. It tests non-setuid programs and setuid programs (setuid root and setuid non-root user). But it doesn't test setgid programs, or `AT_SYMLINK_NOFOLLOW`, or supplementary groups, or `F_OK`. But those tests can be added later.

This pull request also adds support for `F_OK` which was silently absent before.

It's not completely correct. See `src/atcalls.c` for details. But it's a considerable improvement.

One huge concern I have with this pull request is that it really needs some setuid test programs. And that requires linking via an absolute path to `lib/libMacportsLegacySupport.dylib` rather than via the `@executable_path`.

I have changed the `-L` option in the `Makefile` from `-Llib` to `-L"$(shell pwd)"/lib` expecting it to link via an absolute path, but it didn't work.

If I use `-L/opt/local/lib`, it links via an absolute path, but if I use `-L/Users/raf/src/macports-legacy-support/lib`, it links via `@executable_path` and refuses to load the library. It don't know why it ignores the absolute path I supplied. I thought it might be because `/opt/local/lib` is only writable by `root`, and `/Users/raf/...` isn't, but building in a rootonly-writable directory hierarchy didn't help.

So if anyone can tell me how to get the linker to use an absolute path during testing, it would be able to run all the tests (when run by `root`).

In its current (temporary) state, running `make test` tests `faccessat` only in a non-setuid non-setgid program. It prints:

    Run "sudo make test_faccessat_setuid" to test faccessat properly

If that command is run, then a setuid root test and a setuid non-root user test happen. However, the test programs don't link against the library-being-tested via `@executable_path`. They explicitly link against the library in `/opt/local/lib`.

This can change to use the library-being-tested as soon as we can fix the linking problem for setuid programs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -295,6 +295,16 @@ test_cmath: test/test_cmath.cc $(ALLHEADERS)
 	$(info 4: $(CXX) -I$(SRCINCDIR) $(CXXFLAGS) -std=c++11 $< -o test/$@_cxx11)
 	@-$(CXX) -I$(SRCINCDIR) $(CXXFLAGS) -std=c++11 $< -o test/$@_cxx11 &> /dev/null && echo "4: c++11 legacy cmath build success (test succeeded)!" || echo "4: c++11 legacy cmath build failure (test failed)!"
 
+# Special clause for testing faccessat in a setuid program.
+# Must be run by root.
+# Assumes there is a _uucp user.
+# Tests setuid _uucp, setuid root, and setgid tty.
+test_faccessat_setuid: test/test_faccessat
+	@test/do_test_faccessat_setuid "$(BUILDDLIBPATH)"
+
+test_faccessat_setuid_msg:
+	@echo 'Run "sudo make test_faccessat_setuid" to test faccessat properly'
+
 $(TESTRUNS): $(TESTRUNPREFIX)%: $(TESTNAMEPREFIX)%
 	$<
 
@@ -325,11 +335,11 @@ install-slib: $(BUILDSLIBPATH)
 	$(MKINSTALLDIRS) $(DESTDIR)$(LIBDIR)
 	$(INSTALL_DATA) $(BUILDSLIBPATH) $(DESTDIR)$(LIBDIR)
 
-test check: $(TESTRUNS) test_cmath
+test check: $(TESTRUNS) test_cmath test_faccessat_setuid_msg
 
 clean:
 	$(RM) $(foreach D,$(SRCDIR) $(TESTDIR),$D/*.o $D/*.o.* $D/*.d)
-	$(RM) $(BUILDDLIBPATH) $(BUILDSLIBPATH) $(BUILDSYSLIBPATH) $(TESTPRGS) test/test_cmath_*
+	$(RM) $(BUILDDLIBPATH) $(BUILDSLIBPATH) $(BUILDSYSLIBPATH) $(TESTPRGS) test/test_cmath_* test/test_faccessat_setuid
 	@$(RMDIR) $(BUILDDLIBDIR) $(BUILDSLIBDIR)
 
 .PHONY: all dlib slib clean check test $(TESTRUNS) test_cmath
--- a/src/atcalls.c
+++ b/src/atcalls.c
@@ -2,6 +2,7 @@
  * atcalls.c
  *
  * Copyright (c) 2013-2015 Apple Inc. All rights reserved.
+ * Copyright (C) 2023 raf <raf@raf.org>
  *
  * @APPLE_LICENSE_HEADER_START@
  *
@@ -94,23 +95,23 @@ fail:
 
 int faccessat(int dirfd, const char *pathname, int mode, int flags)
 {
-    struct attrlist al = {
-        .bitmapcount = ATTR_BIT_MAP_COUNT,
-        .commonattr  = ATTR_CMN_USERACCESS,
-    };
-    struct {
-        uint32_t length;
-        uint32_t access;
-    } __attribute__((aligned(4), packed)) ab;
-    unsigned long opts = 0;
-
-    ERR_ON(EINVAL, flags & ~AT_SYMLINK_NOFOLLOW);
-    if (flags & AT_SYMLINK_NOFOLLOW)
-        opts |= FSOPT_NOFOLLOW;
-    if (getattrlistat(dirfd, pathname, &al, &ab, sizeof(ab), opts) < 0)
-        return -1;
-    ERR_ON(EACCES, (ab.access & mode) != (uint32_t)mode);
-    return 0;
+    ERR_ON(EINVAL, flags & ~AT_EACCESS);
+    uid_t ruid = getuid(), euid = geteuid();
+    gid_t rgid = getgid(), egid = getegid();
+    int check_euid = ruid != euid && (flags & AT_EACCESS);
+    int check_egid = rgid != egid && (flags & AT_EACCESS);
+    if (check_euid)
+        setreuid(euid, ruid);
+    if (check_egid)
+        setregid(egid, rgid);
+    int access_rc = ATCALL(dirfd, pathname, access(pathname, mode));
+    int access_errno = errno;
+    if (check_euid)
+        setreuid(ruid, euid);
+    if (check_egid)
+        setregid(rgid, egid);
+    errno = access_errno;
+    return access_rc;
 }
 
 #if 0
@@ -299,3 +300,5 @@ int unlinkat(int dirfd, const char *pathname, int flags)
 }
 
 #endif  /* __MP_LEGACY_SUPPORT_ATCALLS__ */
+
+/* vi:set et ts=4 sw=4: */
--- a/test/test_time.cpp
+++ b/test/test_time.cpp
@@ -72,20 +72,6 @@ static void* test_blackhole_thread(void *arg)
   return NULL;
 }
 
-static void* test_wait_thread(void *arg)
-{
-  int c = 0;
-  while ( ++c < 10 )
-  {
-    usleep(100);
-    ::pthread_mutex_lock(&lock);
-    std::cout << "[t2] CLOCK_THREAD_CPUTIME_ID ("<< CLOCK_THREAD_CPUTIME_ID << ") " << time(CLOCK_THREAD_CPUTIME_ID) << std::endl;
-    ::pthread_mutex_unlock(&lock);
-  }
-
-  return NULL;
-}
-
 int main()
 {
   {
--- a/test/do_test_faccessat_setuid
+++ b/test/do_test_faccessat_setuid
@@ -1,0 +1,184 @@
+#!/bin/sh
+
+# Copyright (C) 2023 raf <raf@raf.org>
+#
+# Permission to use, copy, modify, and distribute this software for any
+# purpose with or without fee is hereby granted, provided that the above
+# copyright notice and this permission notice appear in all copies.
+#
+# THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+# WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+# ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+# WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+# ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+# OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+# When run as root, test faccessat() properly
+
+if [ "$(whoami)" != root ]
+then
+	echo 'Run "sudo make test_faccessat_setuid" to test faccessat properly'
+	exit 1
+fi
+
+# Define some names and functions
+
+lib="$1"
+s=test/test_faccessat_setuid
+t=test/tmp
+fail=0
+
+die() { echo "$0: $@" >&2; rm -rf $s $t; exit 1; }
+
+setid() # usage: setid owner group mode fname [createcmd]
+{
+	if [ -n "$5" ]; then $5 $4 || die "$5 $4 failed"; fi
+	chown $1 $4 || die "chown $1 $4 failed"
+	chgrp $2 $4 || die "chgrp $2 $4 failed"
+	chmod $3 $4 || die "chmod $3 $4 failed"
+}
+
+setup() # usage: setup owner group mode
+{
+	cp test/test_faccessat $s || die "cp $s failed"
+	# The following replaces the @executable_path-based library reference
+	# with an absolute path reference to the library-being-tested.
+	# This is needed for a setuid/setgid program to be allowed to load a library.
+	install_name_tool -change "@executable_path/../$lib" "$(pwd)/$lib" $s || die "install_name_tool $s failed"
+	setid $1 $2 $3 $s
+	setid $1 $2 $3 $t mkdir
+}
+
+check()
+{
+	"$@" || fail=1
+}
+
+clean()
+{
+	rm -rf $s $t
+}
+
+get_group()
+{
+	echo $(id $1 | sed -E 's/^.*gid=-?[0-9]+\(//; s/\).*$//')
+}
+
+get_supp_group()
+{
+	echo $(id $1 | sed -E 's/^.*groups=-?[0-9]+\([^)]+\),-?[0-9]+\(//; s/\).*$//')
+}
+
+uucp_group=$(get_group _uucp)
+nobody_group=$(get_group nobody)
+nobody_supp_group=$(get_supp_group nobody)
+
+# Run normal test as setuid _uucp (to test AT_EACCESS)
+
+echo setuid _uucp
+setup _uucp $uucp_group 4755
+check sudo -u nobody $s
+clean
+
+# Run normal test as setuid root (to test AT_EACCESS)
+
+echo setuid root
+setup root wheel 4755
+check sudo -u nobody $s
+clean
+
+# Test different numbers of leading dirs and leading dirs with different
+# permissions (to test leading executable check)
+
+setup _uucp $uucp_group 4755
+setid _uucp $uucp_group 644 $t/f touch
+setid _uucp $uucp_group 755 $t/d1 mkdir
+setid _uucp $uucp_group 000 $t/d2 mkdir
+setid _uucp $uucp_group 644 $t/d1/f touch
+setid _uucp $uucp_group 644 $t/d2/f touch
+
+echo leading dirs ruid=nobody euid=_uucp
+check sudo -u nobody $s test test/ \
+	$t $t/ $t/f \
+	$t/d1 $t/d1/ $t/d1/f \
+	$t/d2 $t/d2/ $t/d2/f
+
+echo leading dirs ruid=root euid=_uucp
+check $s test test/ \
+	$t $t/ $t/f \
+	$t/d1 $t/d1/ $t/d1/f \
+	$t/d2 $t/d2/ $t/d2/f
+
+chmod 755 $s
+
+echo leading dirs nobody
+check sudo -u nobody $s test test/ \
+	$t $t/ $t/f \
+	$t/d1 $t/d1/ $t/d1/f \
+	$t/d2 $t/d2/ $t/d2/f
+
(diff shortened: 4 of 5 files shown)


catap:
Amazing work! I suggest to add all contributors via Co-authored-by

@cjones051073, @jmroot may you review this? It is definitely must to have.

macportsraf:
> Amazing work! I suggest to add all contributors via Co-authored-by

Thanks. I've added co-authors.

src/atcalls.c
```diff
@@ -94,22 +96,105 @@ _fullpathat(int dirfd, const char *relative, char *buf)
...
+
+    struct stat st;
+    if (fstatat(dirfd, pathname, &st, flags & AT_SYMLINK_NOFOLLOW) < 0)
         return -1;
-    ERR_ON(EACCES, (ab.access & mode) != (uint32_t)mode);
+
+    // Check F_OK (fstatat above succeeded, so the file exists)
+
+    if (mode == F_OK)
+        return 0;
```
catap, line 134: This is based on assumption: if `stat` works => process has access to it and it's ok.

I not sure that it's true for `euid`/`egid` case.

Maybe move it after:
```
ERR_ON(EACCES, (st.st_mode & eperms) != eperms)
```
to ensure all checks?

src/atcalls.c
```diff
@@ -94,22 +96,105 @@ _fullpathat(int dirfd, const char *relative, char *buf)
...
+
+        if (mode & R_OK)
+            eperms |= S_IROTH;
+        if (mode & W_OK)
+            eperms |= S_IWOTH;
+        if (mode & X_OK)
+            eperms |= S_IXOTH;
+    }
+
+    ERR_ON(EACCES, (st.st_mode & eperms) != eperms)
```
catap, line 197: Forgotten `;`

src/atcalls.c
macportsraf: No. It's based on the assumption that if stat works, then the file exists. There is no assumption about the (`rwx`) access to the file. Note that `faccessat` can be called with `mode` equal to `F_OK` or a bitmask of the access bits. They can't be combined. So if `F_OK` is specified, there are no access checks to be performed.

However, as explained in the comment above that code, if the `pathname` contains any `/` characters, then it is always the effective users' permissions that determines access to the directories that lead to the file, not the real user's permissions. And although I do think that's incorrect, and that a real system call could do it better, I'm not sure that it's worth fixing that aspect. As I said in the comment, I think that it makes most sense to use `at` functions with `dirfd` referring to the directory that directly contains the file, and `pathname` would not normally have any '/' characters. When that's the case, it's the effective user's permissions that apply to navigating to the directory in question anyway. So I think that in most cases (i.e. when using `at` functions in order to avoid path-based race conditions) this behaviour is indistinguishable from correct behaviour.

src/atcalls.c
macportsraf: Thanks. It's not technically necessary, but it looks better. It's there now.

src/atcalls.c
macportsraf: Actually, it doesn't make sense to knowingly leave it buggy in some cases. I was just being lazy. I'll make it work properly when `pathname` contains `/` characters.

macportsraf:
I've found a big bug:

    ERR_ON(EACCES, (st.st_mode & eperms) != eperms);

That was modelled on the original:

    ERR_ON(EACCES, (ab.access & mode) != (uint32_t)mode);

But the original was matching a bitmask of `R_OK`, `W_OK`, and `X_OK`, to another such bitmask. It was relying on `getattrlistat` having enumerated which access was granted. `faccessat` needs to do that logic itself.

It goes wrong in the case where the target uid and gid don't match, but a supplementary gid does match, because there are two bits set in `eperms`, but only one of them needs to be present in the file permissions (`st.st_mode`). So equality isn't the right test.

So we construct separate file mode bitmasks for read, write, and execute to enumerate what permissions need to be granted, and then check them like this:

    ERR_ON(EACCES, (mode & R_OK) && (st.st_mode & rgranted) == 0);
    ERR_ON(EACCES, (mode & W_OK) && (st.st_mode & wgranted) == 0);
    ERR_ON(EACCES, (mode & X_OK) && (st.st_mode & xgranted) == 0);

Now that I think about it, rather than assume that the overall access algorithm is right, I'll create lots of test files, and make sure that the results match the results of `access`.

catap:
@macportsraf on which system have you test it? I think that you should compare behaviour between something before 10.10 (this implementation) and after 10.10 (apple's one)

macportsraf:
I can test it on 10.14 and 10.6 but when not setuid or setgid.

Until someone finds out how to link using an absolute path to the library-being-tested (rather than via `@executable_path`), I can't test it properly for setuid/setgid programs.

But testing the overall access algorithm can be tested without setuid/setgid. It only needs root access to setup the test. And comparing results against those of `access` is a good idea on any system. That wouldn't change between versions.

catap:
> Until someone finds out how to link using an absolute path to the library-being-tested (rather than via `@executable_path`), I can't test it properly for setuid/setgid programs.

Maybe I'm lost but it seems that you need something like:
```
install_name_tool -change @executable_path/../lib.dylib /absolute/path/to/lib/lib.dylib application
```

Am I lost?

macportsraf:
Thanks! That's great. The setuid tests can now load the library-being-tested. Yay!

I hope it's safe to assume that `install_name_tool` will always be present. On my 10.14 system, it's in `/usr/bin`, and on my 10.6 system, it's in `/opt/local/bin` and `/usr/bin`. So that's encouraging.

And I've moved the ugly shell from the `Makefile` into a separate script where it's less ugly (`test/do_test_faccessat_setuid`).

I still need to add lots of new tests to make sure that everything works, but it's looking good so far.

kencu:
> > Until someone finds out how to link using an absolute path to the library-being-tested (rather than via `@executable_path`), I can't test it properly for setuid/setgid programs.
> 
> Maybe I'm lost but it seems that you need something like:
> 
> ```
> install_name_tool -change @executable_path/../lib.dylib /absolute/path/to/lib/lib.dylib application
> ```
> 
> Am I lost?

The Makefile already does the right thing, using executable_path during building and testing as it should, and then rewriting the install name to an absolute path on installation, as it should.

People who knew what they were doing wrote it.

Don’t break it now…

macportsraf:
I don't doubt it, but the requirements are different now. Now there is a test that needs to be run setuid/setgid before the library installation occurs. So one of the test binaries needs to link using an absolute path to the library-to-be-tested.

But don't worry. The `Makefile` is back the way it was. I had tried to change it to link all the tests with an absolute path but it had no effect, so I put it back. Now the individual test that needs it uses `install_name_tool` to change its linking. That should be harmless.

Makefile
```diff
@@ -295,6 +295,16 @@ test_cmath: test/test_cmath.cc $(ALLHEADERS)
 	$(info 4: $(CXX) -I$(SRCINCDIR) $(CXXFLAGS) -std=c++11 $< -o test/$@_cxx11)
 	@-$(CXX) -I$(SRCINCDIR) $(CXXFLAGS) -std=c++11 $< -o test/$@_cxx11 &> /dev/null && echo "4: c++11 legacy cmath build success (test succeeded)!" || echo "4: c++11 legacy cmath build failure (test failed)!"
 
+# Special clause for testing faccessat in a setuid program.
+# Only counts as a test when run by root.
+# Assumes there is a _uucp user.
+# Tests setuid _uucp, setuid root, and setgid tty.
+test_faccessat_setuid: test/test_faccessat /opt/local/lib/libMacportsLegacySupport.dylib
```
catap, line 302: Why you haven't use `realpath` and `$BUILDSYSLIBPATH`? It allows to test against library which is build in the repo.

macportsraf: Thanks for spotting that. The `/opt/local/lib/libMacportsLegacySupport.dylib` is no longer needed, now that the setuid test can test the library-being-tested rather than an irrelevant already-installed-library.

macportsraf:
Is there an Apple manual entry for `faccessat(2)`? Searching for it on developer.apple.com finds nothing.

I'm adding tests and discovered that `faccessat` on 10.14 doesn't support `AT_SYMLINK_NOFOLLOW`. It returns `-1` with `errno == EINVAL`. The tests pass with this implementation on 10.6 but maybe it shouldn't.

I based the implementation on Linux's manual entry where the valid flags are `AT_EACCESS` and `AT_SYMLINK_NOFOLLOW`, and it implies that it conforms to *POSIX.1-2008*.

The closest I can find to an Apple version of the manpage is FreeBSD's (https://man.freebsd.org/cgi/man.cgi?query=faccessat&sektion=2&n=1) where the valid flags are `AT_EACCESS`, `AT_RESOLVE_BENEATH`, and `AT_EMPTY_PATH`. It claims that it conforms to *The Open Group Extended API Set 2 specification*.

Google find 12 hits for `AT_RESOLVE_BENEATH` so it can probably be ignored. It finds 12900 hits for `AT_EMPTY_PATH`.

Actually, looking at the code at https://github.com/apple-open-source/macos/blob/10.10/xnu/bsd/vfs/vfs_syscalls.c#L5130-L5219 Apple only supports `AT_EACCESS`.

So I'll remove support for `AT_SYMLINK_NOFOLLOW`.

macportsraf:
The testing is done. It's all working now.

The automatic (`make test`) tests aren't enough because there's no `setuid`/`setgid` tests done there, but they're not bad. To really test it requires manually running `sudo make test_faccessat_setuid`. There's a message to that effect at the end of the output of `make test`.

It's behaviour is compatible with `access()` on 10.6, but I think there have been changes to which `errno` value is set in some cases. On 10.14, there's a case where `errno` is set to `ENOENT`, but on 10.6, it's set to `EPERM`. I don't know what version of macOS it changed in. The tests won't pick up on the change because it accepts both. But if it changed before the introduction of `faccessat`, it means that there would be a slight difference on some versions of macOS. If it changed with the introduction of `faccessat`, it should always match.

If you like, I could get the test program to output which value `errno` is being set to in that case, then we(?) can test it on all old versions of macOS to find out when it changed, and then I can make it match based on the macOS version.

Also, thanks to the thorough testing, I realised that I sometimes needed to switch real and effective user ids around calls to `fstatat`. That made me realise that I could probably just modify the original version of `faccessat` to do the same instead (and add support for `F_OK`). I tried the following:

    int faccessat(int dirfd, const char *pathname, int mode, int flags)
    {
        struct attrlist al = {
            .bitmapcount = ATTR_BIT_MAP_COUNT,
            .commonattr  = ATTR_CMN_USERACCESS,
        };
        struct {
            uint32_t length;
            uint32_t access;
        } __attribute__((aligned(4), packed)) ab;
        unsigned long opts = 0;
        ERR_ON(EINVAL, flags & ~AT_EACCESS);
        uid_t ruid = getuid(), euid = geteuid();
        int use_ruid = ruid != euid && !(flags & AT_EACCESS);
        if (mode == F_OK)
        {
            if (use_ruid)
                setreuid(euid, ruid);
            struct stat st;
            int fstatat_rc = fstatat(dirfd, pathname, &st, 0);
            int fstatat_errno = errno;
            if (use_ruid)
                setreuid(ruid, euid);
            ERR_ON(fstatat_errno, fstatat_rc < 0);
        }
        else
        {
            if (use_ruid)
                setreuid(euid, ruid);
            int getattrlistat_rc = getattrlistat(dirfd, pathname, &al, &ab, sizeof(ab), opts);
            int getattrlistat_errno = errno;
            if (use_ruid)
                setreuid(ruid, euid);
            ERR_ON(getattrlistat_errno, getattrlistat_rc < 0);
        }
        ERR_ON(EACCES, (ab.access & mode) != (uint32_t)mode);
        return 0;
    }

And it mostly worked except that there were cases where the `errno` (from `getattrlistat`) was different to the `errno` from `access`, and I don't think I could change it to match.

macportsraf:
Actually, since the only flag that Apple's `faccessat` supports is `AT_EACESS`, I don't see why we can't just use:

    ATCALL(dirfd, pathname, access(pathname, mode))

surrounded by real/effective user id swapping when `AT_EACCESS` is supplied.

I'll try a version that does that.

macportsraf:
The new `setreuid`+`ATCALL`+`setreuid` version works great (only tested on 10.6). And now it's only 9 lines instead of 189 lines. And it's sure to be `errno`-compatible with `access`. And it might even support access control lists now as well.

test/test_time.cpp
```diff
@@ -72,6 +72,7 @@ static void* test_blackhole_thread(void *arg)
   return NULL;
 }
 
+#if 0
```
mascguy, line 75: Is this intended for inclusion with this PR? Or was this a local change, that wasn't supposed to be included?

CC: @catap

catap: as far as I understand this PR it shouldn't affect that. If it affects => somethings is very wrong

src/atcalls.c
```diff
@@ -94,23 +95,23 @@ _fullpathat(int dirfd, const char *relative, char *buf)
...
+    if (check_egid)
+        setregid(egid, rgid);
+    int access_rc = ATCALL(dirfd, pathname, access(pathname, mode));
+    int access_errno = errno;
+    if (check_euid)
+        setreuid(ruid, euid);
+    if (check_egid)
+        setregid(rgid, egid);
+        errno = access_errno;
+    return access_rc;
```
catap, line 114: Interesting idea to emulate that kernel does.

Any way a few remarks.
1. You should respect return code of `setreuid` / `setregid`;
2. You should close a descriptor which is returned by `dirfd`;
3. You may simplify condition like `if (check_euid)` to `if (ruid != euid)` because `flags` should always have `AT_EACCESS`.

test/test_time.cpp
macportsraf: It was intended but unrelated. That function isn't called and caused a warning to that effect when compiling tests, so I #if'd it out. I can restore the warning if you prefer.

mascguy: > It was intended but unrelated. That function isn't called and caused a warning to that effect when compiling tests, so I #if'd it out. I can restore the warning if you prefer.

No, you don't need to do that. But if it's unused, should we simply remove it?

src/atcalls.c
macportsraf: 1. I disagree. According to the manual entries for `setreuid` and `setregid`, they won't fail when swapping real and effective ids. It'll be much uglier and serve no purpose. But I'll change it if that doesn't convince you.
2. `dirfd` in this instance isn't the `dirfd` function that returns a file descriptor from a `DIR` object. It's the name of the first parameter to the `faccessat` function and it should not be closed.
3. That's not true. `flags` can equal 0.

macportsraf: Actually, (1) wouldn't be too ugly. We could return `-1` if the first `setreuid` failed, and `abort` if any of the other three calls failed. It's the logic to undo previous successful calls before returning `-1` that would make it ugly (and handling subsequent hypothetical failures while attempting to do that). I don't like the idea of having code in there to abort, but it's technically OK because it can never be executed, and it's the only appropriate response to such a failure. We can't have any possibility of returning with uids/gids different to what they were upon entry to the function (since we're talking about the possibility of these calls failing). But having calls to `abort` in there could be misleading to anyone reading the code in the future.

It would look like this:

    int faccessat(int dirfd, const char *pathname, int mode, int flags)
    {
        ERR_ON(EINVAL, flags & ~AT_EACCESS);
        uid_t ruid = getuid(), euid = geteuid();
        gid_t rgid = getgid(), egid = getegid();
        int check_euid = ruid != euid && (flags & AT_EACCESS);
        int check_egid = rgid != egid && (flags & AT_EACCESS);
        if (check_euid)
            if (setreuid(euid, ruid) < 0)
                return -1; // This can't happen
        if (check_egid)
            if (setregid(egid, rgid) < 0)
                abort(); // This can't happen
        int access_rc = ATCALL(dirfd, pathname, access(pathname, mode));
        int access_errno = errno;
        if (check_euid)
            if (setreuid(ruid, euid) < 0)
                abort(); // This can't happen
        if (check_egid)
            if (setregid(rgid, egid) < 0)
                abort(); // This can't happen
            errno = access_errno;
        return access_rc;
    }

I don't like this, but I'll do it if you insist.

test/test_time.cpp
macportsraf: Sure. I thought that might be too drastic. It was added last month but not used then. I've removed it. If it needs to be brought back, it's still in the history.

src/atcalls.c
catap: Sorry, I do have a lot of background stress and issues and I was complely worng about `dirfd`; yes, you're right it isn't syscall name. I was confused :)

I love your last approach, but may I ask you to fix formating for `errno = access_errno;` part to avoid one level of tabulation which misslead at reading.

mascguy (approved):

mascguy:
Again, amazing work and collaboration for this PR as well. Thank you all so very much!

src/atcalls.c
macportsraf: Thanks. That's fixed.

cjones051073:
Dear all. It appears the latest changes here have not been properly tested, as they are failing on OSX10.9 and older in the build bots. See e.g.

https://build.macports.org/builders/ports-10.9_x86_64-builder/builds/246086/steps/install-port/logs/stdio

I would roll the released version back to 1.0.13 but leave the devel version broken.

Please take a look and address a.s.a.p.

catap:
@cjones051073 I really doubt that this is a cause.

A log says:
```
/opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_devel_legacy-support/legacy-support-devel/work/compwrap/cc/usr/bin/clang -arch x86_64 -c -Iinclude -Os -fPIC src/atcalls.c -o src/atcalls.dl.o
In file included from src/atcalls.c:59:
include/dirent.h:67:36: error: expected function body after function declarator
int __mpls_libc_closedir(DIR *dir) __MPLS_ALIAS(closedir);
                                   ^
src/atcalls.c:167:40: warning: 'lstat64' is deprecated: first deprecated in OS X 10.6 [-Wdeprecated-declarations]
        return ATCALL(dirfd, pathname, lstat64(pathname, buf));
                                       ^
```

And it seems like a strong pointer into https://github.com/macports/macports-legacy-support/pull/62

mascguy:
> Dear all. It appears the latest changes here have not been properly tested, as they are failing on OSX10.9 and older in the build bots. See e.g.

Chris, I wasn't planning to release this into the wild, until we had at least a week or two to test in the wild via `legacy-support-devel`.

Even then, I was also planning to test building locally across-the-board.

That was stated in one of my messages, from two days ago.

barracuda156:
> @cjones051073 I really doubt that this is a cause.
> 
> A log says:
> 
> ```
> /opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_devel_legacy-support/legacy-support-devel/work/compwrap/cc/usr/bin/clang -arch x86_64 -c -Iinclude -Os -fPIC src/atcalls.c -o src/atcalls.dl.o
> In file included from src/atcalls.c:59:
> include/dirent.h:67:36: error: expected function body after function declarator
> int __mpls_libc_closedir(DIR *dir) __MPLS_ALIAS(closedir);
>                                    ^
> src/atcalls.c:167:40: warning: 'lstat64' is deprecated: first deprecated in OS X 10.6 [-Wdeprecated-declarations]
>         return ATCALL(dirfd, pathname, lstat64(pathname, buf));
>                                        ^
> ```
> 
> And it seems like a strong pointer into #62

This should be fixed now, perhaps?

macportsraf:
> This should be fixed now with https://github.com/macports/macports-legacy-support/commit/707aab751ce95cb394dc09593a5027e97473740e perhaps?

Yes. There was a last minute change to use new macro names (e.g. `__MPLS_ALIAS_I` and `__MPLS_INODE64`) rather than define missing ones (`__DARWIN_ALIAS_I` and `__DARWIN_INODE64`). For consistency, I also used `__MPLS_ALIAS` instead of `__DARWIN_ALIAS` but I forgot to define it. It's fixed by https://github.com/macports/macports-legacy-support/pull/67

PR #67, opened 2023-09-14
Fix dirent.h clang build failures

dpogue:
Does this fix the issues arising from #62? I have no way to test this locally 😞

### DIFF
--- a/include/dirent.h
+++ b/include/dirent.h
@@ -47,18 +47,26 @@
 #undef seekdir
 #undef telldir
 
-/* Fallback to __DARWIN_ALIAS if the other variants are not defined (?) */
-/* Note: I don't know if this makes sense */
-#ifdef __DARWIN_ALIAS_I
-#  define __MPLS_ALIAS_I(sym) __DARWIN_ALIAS_I(sym)
-#else
-#  define __MPLS_ALIAS_I(sym) __DARWIN_ALIAS(sym)
+#ifndef __MPLS_ALIAS
+#  define __MPLS_ALIAS(sym) __DARWIN_ALIAS(sym)
 #endif
 
-#ifdef __DARWIN_INODE64
-#  define __MPLS_INODE64(sym) __DARWIN_INODE64(sym)
-#else
-#  define __MPLS_INODE64(sym) __DARWIN_ALIAS(sym)
+/* Fallback to __DARWIN_ALIAS if the other variants are not defined (?) */
+/* Note: I don't know if this makes sense */
+#ifndef __MPLS_ALIAS_I
+#  ifdef __DARWIN_ALIAS_I
+#    define __MPLS_ALIAS_I(sym) __DARWIN_ALIAS_I(sym)
+#  else
+#    define __MPLS_ALIAS_I(sym) __DARWIN_ALIAS(sym)
+#  endif
+#endif
+
+#ifndef __MPLS_INODE64
+#  ifdef __DARWIN_INODE64
+#    define __MPLS_INODE64(sym) __DARWIN_INODE64(sym)
+#  else
+#    define __MPLS_INODE64(sym) __DARWIN_ALIAS(sym)
+#  endif
 #endif
 
 /* Declare alternative names for the underlying functions for use by the wrappers */


barracuda156:
@catap Kirill, could you try this with Clang on 10.5–10.6, please?

catap:
@barracuda156 do you mean try to rebuild clang? Not right now :(

barracuda156:
> @barracuda156 do you mean try to rebuild clang? Not right now :(

@catap No, just `legacy-support-devel`.

cjones051073:
No. As a first test just build legacy-support on OSX 10.6 through 10.9 using the defsult compilers macports uses to build on these platforms.

catap:
@cjones051073 I may do it in 20 minutes or something like that

include/dirent.h
```diff
@@ -47,6 +47,8 @@
 #undef seekdir
 #undef telldir
 
+#define __MPLS_ALIAS(sym) __DARWIN_ALIAS(sym)
```
cjones051073, line 50: Before defining __MPLS_ALIAS please check it isn't already defined.

cjones051073 (changes requested):

barracuda156:
> @barracuda156 as far as I understand the issue: you need to build something with required headers

@catap Immediate issue is that Clangs failed to build the latest `legacy-support-devel`.

We would still need to verify that a few ports which were broken earlier and got fixed with the latest `legacy-support-devel` with GCC still build with GCC and got fixed with Clang, but that will be step two.

catap:
I've used as local root https://github.com/macports/macports-ports/commit/d87b8a2127f017bdd325f80e7f05e6f4b70d7e94 with patch:
```diff
diff --git a/devel/legacy-support/Portfile b/devel/legacy-support/Portfile
index 28e673e576f..cb46a342f12 100644
--- a/devel/legacy-support/Portfile
+++ b/devel/legacy-support/Portfile
@@ -36,12 +36,12 @@ subport ${name} {
 
 subport ${name}-devel {
     conflicts           ${name}
-    github.setup        macports macports-legacy-support 8b5a02f8129869dbb715a953551f808fbf13d3b0
+    github.setup        dpogue macports-legacy-support fe706f495afc853884f6d15569540d723f854823
     version             20230914
-    revision            0
+    revision            1
     livecheck.type      none
-    checksums           rmd160  ab3a22983d2bf9e7dea8a08491d6bc0c1bc8e35d \
-                        sha256  4603e940dc3d18d93092c71a27d5d02c9f4da634a6efb9c8a1f7f00fc9446b45 \
+    checksums           rmd160  d934403e787d6a5d8d1b5b055da02901c6fc0450 \
+                        sha256  ee07d8f2630dd56ce298a424b9903244a2eb5b71d576b3f6f98cb26454cb8090 \
                         size    72226
     set v_split         [split ${release_ver} .]
     set release_ver     [lindex ${v_split} 0].[lindex ${v_split} 1].99
```

I've run `sudo port deactivate active; sudo port -N install --unrequested legacy-support-devel`. It was succefully installed on: 10.5, 10.6, 10.7, 10.8 and 10.9

catap:
@cjones051073 seems that it fixed issue with isntall. But I haven't dig deeper.

cjones051073:
Thanks. I require my comment above to be fixed, i.e. the define need to check first it doesn't already exist, but once that is in place and tested to show it works I guess we can commit and try again via legacy-support-devel

dpogue:
> i.e. the define need to check first it doesn't already exist

@cjones051073 Should the other defines be wrapped in the same checks? Currently `__MPLS_ALIAS_I` and `__MPLS_INODE64` are defined unconditionally (but with different values based on the existence of other definitions)

cjones051073:
I have no idea.... you need to determine this yourself.

barracuda156:
> @cjones051073 Should the other defines be wrapped in the same checks? Currently `__MPLS_ALIAS_I` and `__MPLS_INODE64` are defined unconditionally (but with different values based on the existence of other definitions)

I guess @macportsraf will know the answer here.

kencu:
Up to now, we decided/required that legacy-support needs to build with the system default compiler on each OS... so for example, for 10.6, that would be /usr/bin/gcc-4.2 

That way, it can be used for other builds such as toolchain parts without concern about a dependency loop.

macportsraf:
I created https://github.com/macports/macports-legacy-support/pull/68 before I knew that this pull request existed. I think https://github.com/macports/macports-legacy-support/pull/68 is a simpler fix.

dpogue:
> I created #68 before I knew that this pull request existed. I think #68 is a simpler fix.

That's what I started with, but the change was rejected because the macro wasn't conditionally defined 😛

macportsraf:
OK. Thanks. Those macros won't be defined anywhere else. We just made up the names three days ago. If they are ever defined elsewhere, I hope they mean the same thing. Your pull request looks good to me. It works for me on 10.6. I'll close my pull request.

PR #68, opened 2023-09-14
Fix recent dirent.h change

macportsraf:
Oops. Sorry about that. That was a last-minute change to avoid defining `__DARWIN_ALIAS_I` and `__DARWIN_INODE64` when not present which might affect others. Instead, it defines `__MPLS_ALIAS_I` and `__MPLS_INODE64`. I forgot to also define `__MPLS_ALIAS` (which I had used for consistency). That's done now and it works on 10.6.

### DIFF
--- a/include/dirent.h
+++ b/include/dirent.h
@@ -61,6 +61,8 @@
 #  define __MPLS_INODE64(sym) __DARWIN_ALIAS(sym)
 #endif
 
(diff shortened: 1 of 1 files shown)


macportsraf:
This has already been fixed with https://github.com/macports/macports-legacy-support/pull/67

PR #69, opened 2023-10-01
Fix dirent.h clang-devel build failure on 10.9, and 10.4 build failure

macportsraf:
Originally, fdopendir was declared in dirent.h
using __DARWIN_ALIAS_I. I think this was needed
because the original implementation had local
variables of type struct statbuf, and so its
translation unit needed to be compiled multiple
times, once for each inode size, and each version
would be differently labelled, and __DARWIN_ALIAS_I
was probably needed to select the correctly
labelled version of fdopendir. But I am guessing.

The current implementation no longer uses any
struct statbuf local variables, and so is only
compiled once. I think this means that it is no
longer necessary to use __DARWIN_ALIAS_I when
declaring fdopendir. Perhaps surprisingly, it
still worked with __DARWIN_ALIAS_I in place.
So maybe I'm guessing wrong.

However, the clang-devel port failed to compile
on macOS-10.9, complaining about the attempt to
overload a function (fdopendir) that differed only
in return type (i.e., LegacySupport's wrapper DIR
struct, as opposed to the system DIR struct). And
maybe I've misunderstood the error message. And
this compiler error didn't appear on other legacy
platforms. So the problem isn't definitely understood.

I might be wrong about why and when __DARWIN_ALIAS_I
is needed, but it seemed that it possibly isn't needed
anymore (if it ever was).

This patch stops declaring fdopendir via __DARWIN_ALIAS_I
and it doesn't break LegacySupport on macOS-10.6 at least.
It compiles and the tests pass. If it works on other legacy
platforms, this should fix clang-devel's build failure by
eliminating the attempt to overload fdopendir.

PR #70, opened 2023-12-18
WIP: time.c: Fix broken CLOCK_MONOTONIC

fhgwright:
Commit bb525c1 replaced the subtly incorrect CLOCK_MONOTONIC fallback
with a version that isn't even monotonic.  This replaces it with a
version that uses the mach SYSTEM_CLOCK, which has been in use in
ntpsec since Jan-2019.

TESTED:
Ran "port destroot" and "port test" on legacy-support-devel, modified
to reference this commit.  Ran on 10.5 ppc, 10.5-10.6 i386, 10.5-10.15
x86_64, and 11.x-14.x arm64.  Built in all cases, and passed tests in
all cases except previously failing ones.
Also ran on 10.4 ppc and 10.4 i386, but using this same patch as
applied to the 1.0.13 release, since the current master doesn't build
on 10.4.  Built and passed tests except for ppc +universal, which
fails to build as before.

### DIFF
--- a/src/time.c
+++ b/src/time.c
@@ -20,9 +20,10 @@
 #if __MP_LEGACY_SUPPORT_GETTIME__
 
 #include <sys/time.h>
-#include <sys/sysctl.h>
 #include <sys/resource.h>
 
+#include <mach/clock.h>
+#include <mach/mach.h>
 #include <mach/mach_init.h>
 #include <mach/mach_port.h>
 #include <mach/mach_time.h>
@@ -46,15 +47,21 @@ int clock_gettime( clockid_t clk_id, struct timespec *ts )
     }
     else if ( CLOCK_MONOTONIC == clk_id )
     {
-      struct timeval boottime;
-      size_t boottime_len = sizeof(boottime);
-      ret = sysctlbyname("kern.boottime", &boottime, &boottime_len, NULL, 0);
-      if (ret != KERN_SUCCESS) { return ret; }
-      struct timeval tv;
-      ret = gettimeofday(&tv, NULL);
-      timersub(&tv, &boottime, &tv);
-      ts->tv_sec  = tv.tv_sec;
-      ts->tv_nsec = tv.tv_usec * 1000;
+      mach_timespec_t mts;
+      static clock_serv_t sclock = 0;
+      /*
+       * Obtain clock port on first call, then reuse it.
+       * Rely on exit cleanup to free it.
+       */
+      if ( 0 == sclock )
+      {
+        mach_port_t mach_host = mach_host_self();
+        host_get_clock_service(mach_host, SYSTEM_CLOCK, &sclock);
+        mach_port_deallocate(mach_task_self(), mach_host);
+      }
+      clock_get_time(sclock, &mts);
+      ts->tv_sec = mts.tv_sec;
+      ts->tv_nsec = mts.tv_nsec;
       ret = 0;
     }
     else if ( CLOCK_PROCESS_CPUTIME_ID == clk_id )


fhgwright:
I'm submitting this just for reference purposes.  I usually bring all my machines and VMs as up to date as possible before running tests, but based on how long this is taking, I don't expect to be able to fully test this PR before tomorrow.  I don't expect the code to change - just the commit message.

This change is orthogonal to #69, so merging order doesn't matter.

Since this is based on the current master, it won't build on 10.4, but there's an alternate version on my `clock-fix-tiger` branch that's based on the 1.0.13 release.  Not intended for a PR, obviously. :-)

fhgwright:
@catap 
As promised.

catap:
Are you sure that it is thread safe?

catap:
About test: can you include an output of unit tests?

fhgwright:
@catap 
> Are you sure that it is thread safe?

In the highly unlikely but not impossible event that two or more threads make the _first_ `CLOCK_MONOTONIC` call at close to the same time, then it can misbehave in a "mostly harmless" fashion.  The result would be that more than one mach port gets allocated while only one gets used, with the maximum number being the number of threads.  Between the low likelihood of this happening and the low cost of the extra ports, it's not clear that it's worth worrying about.

That problem could be _almost_ eliminated by rechecking `s_clock` right before storing into it, and freeing the new port if one had already been stored.  That reduces the effective collision window to a few machine instructions.

The race could be completely avoided by storing `s_clock` with an atomic exchange, and freeing the old port if nonzero, but I don't think there's a way to do an atomic exchange without imposing compiler restrictions.

@catap 
> About test: can you include an output of unit tests?

Like this?
```
/opt/local/var/macports/build/_Volumes_MacPro-Misc_GitHub_.macports_ports_devel_legacy-support/legacy-support-devel/work/compwrap/cxx/usr/bin/clang++ -arch x86_64 -c -Iinclude -Os -stdlib=libc++ test/test_time.cpp -o test/test_time.o
/opt/local/var/macports/build/_Volumes_MacPro-Misc_GitHub_.macports_ports_devel_legacy-support/legacy-support-devel/work/compwrap/cxx/usr/bin/clang++ -arch x86_64 -Llib -L/opt/local/lib -Wl,-headerpad_max_install_names test/test_time.o -lMacportsLegacySupport -o test/test_time
test/test_time
0 0 1000
CLOCK_REALTIME (0)      0 1703038618865310000
CLOCK_REALTIME (0)      0 1703038618865317000
CLOCK_REALTIME (0)      0 1703038618865323000
CLOCK_REALTIME (0)      0 1703038618865328000
CLOCK_REALTIME (0)      0 1703038618865333000
CLOCK_REALTIME (0)      0 1703038618865338000
CLOCK_REALTIME (0)      0 1703038618865343000
CLOCK_REALTIME (0)      0 1703038618865348000
CLOCK_REALTIME (0)      0 1703038618865353000
0 0 1000
CLOCK_MONOTONIC (6)     0 23452108900445
CLOCK_MONOTONIC (6)     0 23452108907128
CLOCK_MONOTONIC (6)     0 23452108913267
CLOCK_MONOTONIC (6)     0 23452108919192
CLOCK_MONOTONIC (6)     0 23452108925057
CLOCK_MONOTONIC (6)     0 23452108930913
CLOCK_MONOTONIC (6)     0 23452108936832
CLOCK_MONOTONIC (6)     0 23452108942681
CLOCK_MONOTONIC (6)     0 23452108948562
0 0 1
CLOCK_MONOTONIC_RAW (4) 0 23452108958287
CLOCK_MONOTONIC_RAW (4) 0 23452108963362
CLOCK_MONOTONIC_RAW (4) 0 23452108968000
CLOCK_MONOTONIC_RAW (4) 0 23452108972600
CLOCK_MONOTONIC_RAW (4) 0 23452108977178
CLOCK_MONOTONIC_RAW (4) 0 23452108981791
CLOCK_MONOTONIC_RAW (4) 0 23452108986368
CLOCK_MONOTONIC_RAW (4) 0 23452108991121
CLOCK_MONOTONIC_RAW (4) 0 23452108995698
0 0 1
CLOCK_MONOTONIC_RAW_APPROX (5) 0 23452109002823
CLOCK_MONOTONIC_RAW_APPROX (5) 0 23452109007918
CLOCK_MONOTONIC_RAW_APPROX (5) 0 23452109012962
CLOCK_MONOTONIC_RAW_APPROX (5) 0 23452109018049
CLOCK_MONOTONIC_RAW_APPROX (5) 0 23452109023089
CLOCK_MONOTONIC_RAW_APPROX (5) 0 23452109028091
CLOCK_MONOTONIC_RAW_APPROX (5) 0 23452109033171
CLOCK_MONOTONIC_RAW_APPROX (5) 0 23452109038194
CLOCK_MONOTONIC_RAW_APPROX (5) 0 23452109043242
0 0 1
CLOCK_UPTIME_RAW (8) 0 23452109049567
CLOCK_UPTIME_RAW (8) 0 23452109053993
CLOCK_UPTIME_RAW (8) 0 23452109058408
CLOCK_UPTIME_RAW (8) 0 23452109062770
CLOCK_UPTIME_RAW (8) 0 23452109067135
CLOCK_UPTIME_RAW (8) 0 23452109071507
CLOCK_UPTIME_RAW (8) 0 23452109075870
CLOCK_UPTIME_RAW (8) 0 23452109080261
CLOCK_UPTIME_RAW (8) 0 23452109084668
0 0 1
CLOCK_UPTIME_RAW_APPROX (8) 0 23452109091674
CLOCK_UPTIME_RAW_APPROX (8) 0 23452109096640
CLOCK_UPTIME_RAW_APPROX (8) 0 23452109101590
CLOCK_UPTIME_RAW_APPROX (8) 0 23452109106454
CLOCK_UPTIME_RAW_APPROX (8) 0 23452109111341
CLOCK_UPTIME_RAW_APPROX (8) 0 23452109116184
CLOCK_UPTIME_RAW_APPROX (8) 0 23452109121030
CLOCK_UPTIME_RAW_APPROX (8) 0 23452109125944
CLOCK_UPTIME_RAW_APPROX (8) 0 23452109130806
0 0 1000
CLOCK_PROCESS_CPUTIME_ID (12) 0 1796000
CLOCK_PROCESS_CPUTIME_ID (12) 0 1802000
CLOCK_PROCESS_CPUTIME_ID (12) 0 1808000
CLOCK_PROCESS_CPUTIME_ID (12) 0 1812000
CLOCK_PROCESS_CPUTIME_ID (12) 0 1817000
CLOCK_PROCESS_CPUTIME_ID (12) 0 1822000
CLOCK_PROCESS_CPUTIME_ID (12) 0 1827000
CLOCK_PROCESS_CPUTIME_ID (12) 0 1832000
CLOCK_PROCESS_CPUTIME_ID (12) 0 1836000
0 0 1000
CLOCK_THREAD_CPUTIME_ID (16) 0 1847000
CLOCK_THREAD_CPUTIME_ID (16) 0 1856000
CLOCK_THREAD_CPUTIME_ID (16) 0 1863000
CLOCK_THREAD_CPUTIME_ID (16) 0 1869000
CLOCK_THREAD_CPUTIME_ID (16) 0 1876000
CLOCK_THREAD_CPUTIME_ID (16) 0 1883000
CLOCK_THREAD_CPUTIME_ID (16) 0 1889000
CLOCK_THREAD_CPUTIME_ID (16) 0 1897000
CLOCK_THREAD_CPUTIME_ID (16) 0 1903000
0 0 0 00  0 [0~10000000] consumed thread CPU time: 13000 ns
0 0 [0~1000] consumed thread CPU time: 31000 ns
[1~10000000] consumed thread CPU time: 4000 ns
0 0 0 [0~10000000 0] consumed thread CPU time: 4000 ns
0 [0 1~0 1000] consumed thread CPU time: 10000 ns
[0 1~10000000] consumed thread CPU ti0m e: 5000 ns
0 [sleep for 01 00us] consumed thread CPU time: 27000 ns
0 [0~1000] consumed thread CPU time: 9000 ns
[0 03 ~10000000] consumed thread CPU time: 5000 ns
[01 ~0 1000] consumed thread CPU time: 5000 ns
[3~10000000] consumed thread CPU time: 0 17000 ns
0 [1~1000] consumed thre0a d CPU time: 0 5000 ns
0 [00~ 10000000] consumed thread CPU time: 01 6000 ns
[sleep for 1000u s0]  consumed thread CPU time: 10000 ns
[3~1000] consumed thread CPU time: 103 000 ns
[1~01 00000000 ] consumed thread CPU time: 8000 ns
[0 4~01 000] consumed thread CPU time: 12000 ns
[0 0~10000000] consumed th0r ead CPU time: 5000 ns
[1~1000] consumed thread CPU time: 4000 ns
0 0 [0~1000] consumed thread CPU time: 2000 ns
0 [sleep for 100us] consumed thread CPU time: 26000 ns
0 0 [sleep for 100us] consumed thread CPU time: 9000 ns
0 0 [sleep for 100us] consumed thread CPU time: 8000 ns
0 0 [sleep for 100us] consumed thread CPU time: 7000 ns
0 0 [sleep for 100us] consumed thread CPU time: 8000 ns
0 0 [sleep for 100us] consumed thread CPU time: 8000 ns
0 0 [sleep for 100us] consumed thread CPU time: 9000 ns
0 [sleep for 100000us] consumed thread CPU time: 27000 ns
0 0 [sleep for 100000us] consumed thread CPU time: 26000 ns
0 0 [sleep for 100000us] consumed thread CPU time: 25000 ns
0 0 [sleep for 100000us] consumed thread CPU time: 23000 ns
0 0 [sleep for 100000us] consumed thread CPU time: 23000 ns
0 0 [sleep for 100000us] consumed thread CPU time: 22000 ns
0 0 [sleep for 100000us] consumed thread CPU time: 23000 ns
0 0 [sleep for 100000us] consumed thread CPU time: 22000 ns
0 0 [sleep for 100000us] consumed thread CPU time: 22000 ns
```

catap:
@fhgwright I've re-read man page for `clock_gettime` on my macOS 12, let me quote:

> `CLOCK_MONOTONIC` clock that increments monotonically, tracking the time since an arbitrary point, and will continue to increment while the system is asleep.
>
> `CLOCK_MONOTONIC_RAW` clock that increments monotonically, tracking the time since an arbitrary point like `CLOCK_MONOTONIC`.  However, this clock is unaffected by frequency or time adjustments.  It should not be compared to other system time sources.

As you may see man page doesn't state that `CLOCK_MONOTONIC` shouldn't be affected by time adjustments. If someone needs something that shouldn't be affected by time adjustment, well... he should use `CLOCK_MONOTONIC_RAW` instead `CLOCK_MONOTONIC`, at least I read that from this man page.

So, let run some tests on your branch. I use two machines: macOS 12 and macOS 10.5 i386.

on macOS 12 `test_time` works as:
```
0 0 1000
CLOCK_MONOTONIC (6)     0 751042320283000
CLOCK_MONOTONIC (6)     0 751042320286000
CLOCK_MONOTONIC (6)     0 751042320288000
CLOCK_MONOTONIC (6)     0 751042320290000
CLOCK_MONOTONIC (6)     0 751042320292000
CLOCK_MONOTONIC (6)     0 751042320294000
CLOCK_MONOTONIC (6)     0 751042320296000
CLOCK_MONOTONIC (6)     0 751042320298000
CLOCK_MONOTONIC (6)     0 751042320301000
0 0 1
CLOCK_MONOTONIC_RAW (4) 0 751034437383231
CLOCK_MONOTONIC_RAW (4) 0 751034437385543
CLOCK_MONOTONIC_RAW (4) 0 751034437387574
CLOCK_MONOTONIC_RAW (4) 0 751034437389614
CLOCK_MONOTONIC_RAW (4) 0 751034437391595
CLOCK_MONOTONIC_RAW (4) 0 751034437393875
CLOCK_MONOTONIC_RAW (4) 0 751034437396117
CLOCK_MONOTONIC_RAW (4) 0 751034437398158
CLOCK_MONOTONIC_RAW (4) 0 751034437400214
```
and on 10.5:
```
0 0 1000
0 CLOCK_MONOTONIC (6)     313309402660
0 CLOCK_MONOTONIC (6)     313309418519
0 CLOCK_MONOTONIC (6)     313309425414
0 CLOCK_MONOTONIC (6)     313309432804
0 CLOCK_MONOTONIC (6)     313309439301
0 CLOCK_MONOTONIC (6)     313309445661
0 CLOCK_MONOTONIC (6)     313309451838
0 CLOCK_MONOTONIC (6)     313309458017
0 CLOCK_MONOTONIC (6)     313309464209
0 0 1
0 CLOCK_MONOTONIC_RAW (4) 313309476702
0 CLOCK_MONOTONIC_RAW (4) 313309482605
0 CLOCK_MONOTONIC_RAW (4) 313309486641
0 CLOCK_MONOTONIC_RAW (4) 313309490429
0 CLOCK_MONOTONIC_RAW (4) 313309494153
0 CLOCK_MONOTONIC_RAW (4) 313309497900
0 CLOCK_MONOTONIC_RAW (4) 313309501787
0 CLOCK_MONOTONIC_RAW (4) 313309505568
0 CLOCK_MONOTONIC_RAW (4) 313309509714
```

as you may see `CLOCK_MONOTONIC` should have resolution `1000` and not `1`, and your changes doesn't follow this limits.

The next test is sleep for couple of minutes. Now it works like this:
```
0 0 1000
0 CLOCK_MONOTONIC (6)     404381008568
0 CLOCK_MONOTONIC (6)     404381018990
0 CLOCK_MONOTONIC (6)     404381025839
0 CLOCK_MONOTONIC (6)     404381032289
0 CLOCK_MONOTONIC (6)     404381038797
0 CLOCK_MONOTONIC (6)     404381045204
0 CLOCK_MONOTONIC (6)     404381051484
0 CLOCK_MONOTONIC (6)     404381057694
0 CLOCK_MONOTONIC (6)     404381063976
0 0 1
0 CLOCK_MONOTONIC_RAW (4) 404381077503
0 CLOCK_MONOTONIC_RAW (4) 404381083343
0 CLOCK_MONOTONIC_RAW (4) 404381088203
0 CLOCK_MONOTONIC_RAW (4) 404381092147
0 CLOCK_MONOTONIC_RAW (4) 404381095990
0 CLOCK_MONOTONIC_RAW (4) 404381099772
0 CLOCK_MONOTONIC_RAW (4) 404381103722
0 CLOCK_MONOTONIC_RAW (4) 404381107731
0 CLOCK_MONOTONIC_RAW (4) 404381111523
```

No big differences between last call `CLOCK_MONOTONIC` and `CLOCK_MONOTONIC_RAW`.

If I switch to master I have:
```
0 0 1000
0 CLOCK_MONOTONIC (6)     537305184000
0 CLOCK_MONOTONIC (6)     537305202000
0 CLOCK_MONOTONIC (6)     537305216000
0 CLOCK_MONOTONIC (6)     537305228000
0 CLOCK_MONOTONIC (6)     537305241000
0 CLOCK_MONOTONIC (6)     537305254000
0 CLOCK_MONOTONIC (6)     537305266000
0 CLOCK_MONOTONIC (6)     537305278000
0 CLOCK_MONOTONIC (6)     537305290000
0 0 1
0 CLOCK_MONOTONIC_RAW (4) 538123353707
0 CLOCK_MONOTONIC_RAW (4) 538123375877
0 CLOCK_MONOTONIC_RAW (4) 538123405869
0 CLOCK_MONOTONIC_RAW (4) 538123534539
0 CLOCK_MONOTONIC_RAW (4) 538123543129
0 CLOCK_MONOTONIC_RAW (4) 538123548970
0 CLOCK_MONOTONIC_RAW (4) 538123554446
0 CLOCK_MONOTONIC_RAW (4) 538123559950
0 CLOCK_MONOTONIC_RAW (4) 538123565752
```

and after sleep the master branch behaves like:
```
0 0 1000
0 CLOCK_MONOTONIC (6)     570271919000
0 CLOCK_MONOTONIC (6)     570271939000
0 CLOCK_MONOTONIC (6)     570271954000
0 CLOCK_MONOTONIC (6)     570271967000
0 CLOCK_MONOTONIC (6)     570271980000
0 CLOCK_MONOTONIC (6)     570271994000
0 CLOCK_MONOTONIC (6)     570272007000
0 CLOCK_MONOTONIC (6)     570272020000
0 CLOCK_MONOTONIC (6)     570272034000
0 0 1
0 CLOCK_MONOTONIC_RAW (4) 571090619898
0 CLOCK_MONOTONIC_RAW (4) 571090625913
0 CLOCK_MONOTONIC_RAW (4) 571090629885
0 CLOCK_MONOTONIC_RAW (4) 571090633842
0 CLOCK_MONOTONIC_RAW (4) 571090637905
0 CLOCK_MONOTONIC_RAW (4) 571090641751
0 CLOCK_MONOTONIC_RAW (4) 571090645795
0 CLOCK_MONOTONIC_RAW (4) 571090649744
0 CLOCK_MONOTONIC_RAW (4) 571090653621
```
differences between larger, but hasn't changed as well.

and as reference I'd like to use macOS 10.15 inside VM to investigate behaviour of that function in apple implementation:

Master branch:
```
0 0 1000
CLOCK_MONOTONIC (6)     0 115876383000
CLOCK_MONOTONIC (6)     0 115876388000
CLOCK_MONOTONIC (6)     0 115876390000
CLOCK_MONOTONIC (6)     0 115876392000
CLOCK_MONOTONIC (6)     0 115876395000
CLOCK_MONOTONIC (6)     0 115876397000
CLOCK_MONOTONIC (6)     0 115876399000
CLOCK_MONOTONIC (6)     0 115876401000
CLOCK_MONOTONIC (6)     0 115876403000
0 0 1
CLOCK_MONOTONIC_RAW (4) 0 116779642332
CLOCK_MONOTONIC_RAW (4) 0 116779644790
CLOCK_MONOTONIC_RAW (4) 0 116779647040
CLOCK_MONOTONIC_RAW (4) 0 116779649234
CLOCK_MONOTONIC_RAW (4) 0 116779651496
CLOCK_MONOTONIC_RAW (4) 0 116779653752
CLOCK_MONOTONIC_RAW (4) 0 116779655949
CLOCK_MONOTONIC_RAW (4) 0 116779658137
CLOCK_MONOTONIC_RAW (4) 0 116779660331
```

Huge differences between the last `CLOCK_MONOTONIC` and the first `CLOCK_MONOTONIC_RAW` is here.

So, my points:
 1. I not sure that your bug is bug at all, from my point of view it seems like expected behaviour which I read in man page for `clock_gettime`;
 2. current implementation behaves like implementation in macOS 10.15;
 3. your code should be fixed to respect resolution.

fhgwright:
@catap 
> @fhgwright I've re-read man page for `clock_gettime` on my macOS 12, let me quote:
> 
> > `CLOCK_MONOTONIC` clock that increments monotonically, tracking the time since an arbitrary point, and will continue to increment while the system is asleep.
> > `CLOCK_MONOTONIC_RAW` clock that increments monotonically, tracking the time since an arbitrary point like `CLOCK_MONOTONIC`.  However, this clock is unaffected by frequency or time adjustments.  It should not be compared to other system time sources.
> 
> As you may see man page doesn't state that `CLOCK_MONOTONIC` shouldn't be affected by time adjustments. If someone needs something that shouldn't be affected by time adjustment, well... he should use `CLOCK_MONOTONIC_RAW` instead `CLOCK_MONOTONIC`, at least I read that from this man page.

But what you're missing is the word `monotonically`.   Time adjustments can be either slewing or step adjustments.  The former are always monotonic, while the latter may not be.  `CLOCK_MONOTONIC` was created for the express purpose of excluding step adjustments.  It does normally include slewing adjustments, and some documentation explicitly states that.  But both `CLOCK_REALTIME` and `gettimeofday()` include step adjustments; hence any timescale derived from them also includes step adjustments, violating the monotonicity requirement.

Apple's code looks functionally similar to yours, though with what appears to be a more complicated treatment of "boottime".  It does use `gettimeofday()`, so it almost certainly suffers from the same monotonicity problem.  If the goal of `legacy-support` is to be bugwise compatible with later OS versions, then I guess the current code is "correct". :-) In that case, it would make sense to include a comment as to why the code is intentionally broken.

I believe it should be possible to create a test that demonstrates the problem, both in `legacy-support` and in Apple's code, though it's not trivial and not something I expect to get to right away.  It's an inherently disruptive (and root-requiring) test, so it couldn't be part of the normal test suite.

[...]

> So, my points:
> 
> 1. I not sure that your bug is bug at all, from my point of view it seems like expected behaviour which I read in man page for `clock_gettime`;

See above.

> 2. current implementation behaves like implementation in macOS 10.15;

Sort of, though if "boottime" is constant it should be cacheable (though the complexity of Apple's `_mach_boottime_usec()` suggests that it isn't constant, which is also weird).  And `sysctlbyname()` is unnecessary slowness compared to `sysctl()`, which in turn is unnecessary slowness compared to `mach_boottime_usec()` if the latter is available.

> 3. your code should be fixed to respect resolution.

Yes, I missed that.  That's because bb525c1 changed the resolution without updating `clock_getres()`, which is in turn because `clock_getres()` was incorrect in bb525c1^, and became stealthily correct in bb525c1. :-)

Some platforms are pretty sloppy with `clock_getres()`, though it looks like Apple's code tries to do the best it can given the integer nanosecond limitation.  Which is more than I can say for `legacy-support`. :-)

catap:
> Apple's code looks functionally similar to yours, though with what appears to be a more complicated treatment of "boottime". It does use `gettimeofday()`, so it almost certainly suffers from the same monotonicity problem. If the goal of `legacy-support` is to be bugwise compatible with later OS versions, then I guess the current code is "correct". :-) In that case, it would make sense to include a comment as to why the code is intentionally broken.

The goal of this library is implement missed function at old OS in the way how it's implemented in modern OS.

If current implementation works like the code from modern OS => this is expected implementation. 

> Sort of, though if "boottime" is constant it should be cacheable (though the complexity of Apple's `_mach_boottime_usec()` suggests that it isn't constant, which is also weird). And `sysctlbyname()` is unnecessary slowness compared to `sysctl()`, which in turn is unnecessary slowness compared to `mach_boottime_usec()` if the latter is available.

Let see the apple code! I've used https://github.com/apple-open-source/macos/blob/14.2/Libc/gen/clock_gettime.c as references.

```c
int
clock_gettime(clockid_t clk_id, struct timespec *tp)
{
    switch(clk_id){
...
    case CLOCK_MONOTONIC: {
        struct timeval tv;
        uint64_t boottime_usec;
        int ret = _mach_boottime_usec(&boottime_usec, &tv);
        struct timeval boottime = {
            .tv_sec = boottime_usec / USEC_PER_SEC,
            .tv_usec = boottime_usec % USEC_PER_SEC
        };
        timersub(&tv, &boottime, &tv);
        TIMEVAL_TO_TIMESPEC(&tv, tp);
        return ret;
    }
...
}

static int
_mach_boottime_usec(uint64_t *boottime, struct timeval *realtime)
{
    uint64_t bt1 = 0, bt2 = 0;
    int ret;
    do {
        bt1 = mach_boottime_usec();
        if (os_slowpath(bt1 == 0)) bt1 = _boottime_fallback_usec();

        atomic_thread_fence(memory_order_seq_cst);

        ret = gettimeofday(realtime, NULL);
        if (ret != 0) return ret;

        atomic_thread_fence(memory_order_seq_cst);

        bt2 = mach_boottime_usec();
        if (os_slowpath(bt2 == 0)) bt2 = _boottime_fallback_usec();
    } while (os_slowpath(bt1 != bt2));
    *boottime = bt1;
    return 0;
}
```

With exception that I haven't use `os_slowpath` current code looks quite similar with Apple's one, isn't it?

catap:
> This replaces it with a version that uses the mach SYSTEM_CLOCK, which has been in use in ntpsec since Jan-2019.

BTW, can I ask you to share how do you use it in ntpsec?

fhgwright:
@catap 
> > Apple's code looks functionally similar to yours, though with what appears to be a more complicated treatment of "boottime". It does use `gettimeofday()`, so it almost certainly suffers from the same monotonicity problem. If the goal of `legacy-support` is to be bugwise compatible with later OS versions, then I guess the current code is "correct". :-) In that case, it would make sense to include a comment as to why the code is intentionally broken.
> 
> The goal of this library is implement missed function at old OS in the way how it's implemented in modern OS.
> 
> If current implementation works like the code from modern OS => this is expected implementation.

True, though it actually doesn't.  I don't yet know whether Apple's code is fully correct or not.  If step adjustments are applied to "boottime" in a manner that cancels the effect on "timeofday", then that takes care of the monotonicity issue.  If not, then their code has a bug.  And of course, since it appears that "boottime" isn't entirely constant, that makes the name a lie.

> > Sort of, though if "boottime" is constant it should be cacheable (though the complexity of Apple's `_mach_boottime_usec()` suggests that it isn't constant, which is also weird). And `sysctlbyname()` is unnecessary slowness compared to `sysctl()`, which in turn is unnecessary slowness compared to `mach_boottime_usec()` if the latter is available.
> 
> Let see the apple code! I've used https://github.com/apple-open-source/macos/blob/14.2/Libc/gen/clock_gettime.c as references.

I'm looking at https://github.com/apple-oss-distributions/Libc/blob/c5a3293354e22262702a3add5b2dfc9bb0b93b85/gen/clock_gettime.c, which is essentially the same.

> With exception that I haven't use `os_slowpath` current code looks quite similar with Apple's one, isn't it?

Not really.  The important part isn't the `os_slowpath`, which I believe is just a branch-prediction hint.  It's the fact that `gettimeofday()` is read in a sandwich between two calls of `mach_boottime_usec()`, repeating the whole sequence if the two values differ.  And I'm not sure what the `_boottime_fallback_usec()` stuff is about - maybe for early startup before the time is known?  Anyway, the idea is to capture a consistent "timeofday/boottime" pair.  Not doing so is a race condition.  This is a much narrower (and harder to reproduce) failure than what I originally suspected, but it's nevertheless a bug.  And of course `sysctlbyname()` is almost certainly _way_ slower than `mach_boottime_usec()`.


> > This replaces it with a version that uses the mach SYSTEM_CLOCK, which has been in use in ntpsec since Jan-2019.
> 
> BTW, can I ask you to share how do you use it in ntpsec?

I just provide fallbacks for `clock_gettime()` and `clock_settime()`.  It doesn't appear that `CLOCK_MONOTONIC` is used by `ntpd` itself, though various test programs use it.

https://gitlab.com/fhgwright/ntpsec/-/blob/mac-fixes/include/ntp_machine.h

catap:
> True, though it actually doesn't. I don't yet know whether Apple's code is fully correct or not. If step adjustments are applied to "boottime" in a manner that cancels the effect on "timeofday", then that takes care of the monotonicity issue. If not, then their code has a bug. And of course, since it appears that "boottime" isn't entirely constant, that makes the name a lie.

See the unit test which explains how it works and that value is expected: https://github.com/apple-open-source/macos/blob/master/xnu/tests/mach_boottime_usec.c

> Not really. The important part isn't the `os_slowpath`, which I believe is just a branch-prediction hint. It's the fact that `gettimeofday()` is read in a sandwich between two calls of `mach_boottime_usec()`, repeating the whole sequence if the two values differ. And I'm not sure what the `_boottime_fallback_usec()` stuff is about - maybe for early startup before the time is known? Anyway, the idea is to capture a consistent "timeofday/boottime" pair. Not doing so is a race condition. This is a much narrower (and harder to reproduce) failure than what I originally suspected, but it's nevertheless a bug. And of course `sysctlbyname()` is almost certainly _way_ slower than `mach_boottime_usec()`.

Let me share the `_boottime_fallback_usec` from that file:
```c
static uint64_t
_boottime_fallback_usec(void)
{
    struct timeval tv;
    size_t len = sizeof(tv);
    int ret = sysctlbyname("kern.boottime", &tv, &len, NULL, 0);
    if (ret == -1) return 0;
    return (uint64_t)tv.tv_sec * USEC_PER_SEC + (uint64_t)tv.tv_usec;
}
```

`mach_boottime_usec` returns near the same value, but much faster.

My implementation was simple go via slow path and without loop.

My goal was to implement it as closed to Apple as possible to make behaviour for software expected. If we changed it's behaviour from Apple-way to something different, we may introduce unexpected bugs in random software.

Let's don't do it.

fhgwright:
> > True, though it actually doesn't. I don't yet know whether Apple's code is fully correct or not. If step adjustments are applied to "boottime" in a manner that cancels the effect on "timeofday", then that takes care of the monotonicity issue. If not, then their code has a bug. And of course, since it appears that "boottime" isn't entirely constant, that makes the name a lie.
> 
> See the unit test which explains how it works and that value is expected: https://github.com/apple-open-source/macos/blob/master/xnu/tests/mach_boottime_usec.c

The unit test isn't designed to simulate the race condition that the loop is there to avoid (which would actually be fairly difficult).

> > Not really. The important part isn't the `os_slowpath`, which I believe is just a branch-prediction hint. It's the fact that `gettimeofday()` is read in a sandwich between two calls of `mach_boottime_usec()`, repeating the whole sequence if the two values differ. And I'm not sure what the `_boottime_fallback_usec()` stuff is about - maybe for early startup before the time is known? Anyway, the idea is to capture a consistent "timeofday/boottime" pair. Not doing so is a race condition. This is a much narrower (and harder to reproduce) failure than what I originally suspected, but it's nevertheless a bug. And of course `sysctlbyname()` is almost certainly _way_ slower than `mach_boottime_usec()`.
> 
> Let me share the `_boottime_fallback_usec` from that file:
> 
> ```c
> static uint64_t
> _boottime_fallback_usec(void)
> {
>     struct timeval tv;
>     size_t len = sizeof(tv);
>     int ret = sysctlbyname("kern.boottime", &tv, &len, NULL, 0);
>     if (ret == -1) return 0;
>     return (uint64_t)tv.tv_sec * USEC_PER_SEC + (uint64_t)tv.tv_usec;
> }
> ```
> 
> `mach_boottime_usec` returns near the same value, but much faster.
> 
> My implementation was simple go via slow path and without loop.

But the loop is absolutely essential to capture a consistent boottime/timeofday pair.  Otherwise there's a race.  And the fallback code is only used when the primary code fails to return a value, which is almost certainly not the normal case.  If it were even remotely common, I'm sure they would have used `sysctl()` rather than `sysctlbyname()`.

> My goal was to implement it as closed to Apple as possible to make behaviour for software expected. If we changed it's behaviour from Apple-way to something different, we may introduce unexpected bugs in random software.

But your code is _not_ "as close to Apple as possible".  If my guess is correct as to why Apple has the complexity of the loop (I'm sure it's not there just because they thought that extra complexity was a plus), then your code has a risk of failing monotonicity, which can cause _very_ nasty bugs in software that assumes that a monotonic timescale is actually monotonic.

> Let's don't do it.

I agree that my code is probably not the best choice here.  I do _not_ agree that your code is correct, or even an improvement over what it replaced.

catap:
> But your code is _not_ "as close to Apple as possible". If my guess is correct as to why Apple has the complexity of the loop (I'm sure it's not there just because they thought that extra complexity was a plus), then your code has a risk of failing monotonicity, which can cause _very_ nasty bugs in software that assumes that a monotonic timescale is actually monotonic.

Apple code quite easy to make quite not monotonic:
1. Select time from tomorrow via `date`;
2. run trivial program which prints differences between two `clock_gettime(CLOCK_MONOTONIC)` with delay 1 minute;
3. when program is wait just run `ntpdate` on another console.

Congrats, the program returns negative value.

Have I missed something?

PR #72, opened 2024-02-18
Fix some expected fdopendir() failures, plus some other cleanups

fhgwright:
This was triggered by noticing that the fdopendir(stdin) test was failing (due to returning the wrong errno) when run from a pipe rather than a tty or pty.  This led to adding two more tests, which in turn led to finding additional bugs.  These bugs are fixed here, and the commit order is rearranged to put the fixes before the test additions, to avoid creating a pothole in the history.

There are also some additional cleanups, both related and unrelated, as separate commits.

See the individual commit messages for more information.

This has been tested on all non-oddball platforms except 13.x+ x86_64, as well as on 10.4 x86_64 (which tends to be a problem case).  Building and testing "bare" in this repo (non-universal only) successfully builds and passes tests on all such platforms.  Building and testing in the `legacy-support-devel` context has some failures, but these are the same failures with and without this PR, and are unrelated to fdopendir().

Tested on:
```
Mac OS X 10.4.11 8S2167, x86_64, Xcode 2.5 8M2558
macOS 12.7.3 21H1015, x86_64, Xcode 14.2 14C18
macOS 12.7.3 21H1015, arm64, Xcode 14.2 14C18
macOS 13.6.4 22G513, arm64, Xcode 15.2 15C500b
macOS 14.3.1 23D60, arm64, Xcode 15.2 15C500b
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 *.o
+*.o.*
 *.d
 *.dylib
 lib
 *~
+*.dSYM
--- a/src/fdopendir.c
+++ b/src/fdopendir.c
@@ -24,6 +24,10 @@
 
 #include <dirent.h>
 
+#if __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1050
+#define __dd_fd dd_fd
+#endif /* __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1050 */
+
 /*
  * Implementation behavior largely follows these man page descriptions:
  *
@@ -51,13 +55,8 @@ DIR *fdopendir(int dirfd) {
      * A subsequent closedir() will close dirfd
      */
 
-    #if __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ == 1040
-    (void)close(dir->dd_fd);
-    dir->dd_fd = dirfd;
-    #else
     (void)close(dir->__dd_fd);
     dir->__dd_fd = dirfd;
-    #endif
 
     /*
      * Rewind to the start of the directory, in case the underlying file
--- a/src/fdopendir.c
+++ b/src/fdopendir.c
@@ -23,6 +23,7 @@
 #include "common-priv.h"
 
 #include <dirent.h>
+#include <sys/stat.h>
 
 #if __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1050
 #define __dd_fd dd_fd
@@ -36,19 +37,26 @@
  */
 
 DIR *fdopendir(int dirfd) {
+    DIR *dir;
+    struct stat dirstat;
 
-    /* Check dirfd here (for macos-10.4, see _ATCALL() and best_fchdir()) */
+    /* Fail if dirfd isn't a valid open fd */
+    if (fstat(dirfd, &dirstat) < 0) {
+        return NULL;
+    }
 
-    if (dirfd < 0) {
-        errno = EBADF;
-        return 0;
+    /* Fail if dirfd isn't a directory */
+    if (!S_ISDIR(dirstat.st_mode)) {
+        errno = ENOTDIR;
+        return NULL;
     }
 
     /* Open given directory fd safely for iteration via readdir */
 
-    DIR *dir = _ATCALL(dirfd, ".", NULL, opendir("."));
-    if (!dir)
-        return 0;
+    dir = _ATCALL(dirfd, ".", NULL, opendir("."));
+    if (!dir) {
+        return NULL;
+    }
 
     /*
      * Replace underlying fd with supplied dirfd
--- a/src/fdopendir.c
+++ b/src/fdopendir.c
@@ -46,7 +46,10 @@ DIR *fdopendir(int dirfd) {
     if (!dir)
         return 0;
 
-    /* Replace underlying fd with equivalent given fd (closed by closedir) */
+    /*
+     * Replace underlying fd with supplied dirfd
+     * A subsequent closedir() will close dirfd
+     */
 
     #if __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ == 1040
     (void)close(dir->dd_fd);
@@ -56,11 +59,14 @@ DIR *fdopendir(int dirfd) {
     dir->__dd_fd = dirfd;
     #endif
 
-    /* Rewind to the start of the directory (in case it's not there already) */
+    /*
+     * Rewind to the start of the directory, in case the underlying file
+     * is not positioned at the start
+     */
 
     rewinddir(dir);
 
-    /* Close given fd on exec (just in case not already done) */
+    /* Close given fd on exec (as per fdopendir() docs) */
 
     (void)fcntl(dirfd, F_SETFD, FD_CLOEXEC);
 
--- a/test/.gitignore
+++ b/test/.gitignore
@@ -1,7 +1,5 @@
-test_arc4random
-test_cmath_cxx03
-test_cmath_cxx11
-test_realpath
-test_sysconf
-test_time
-test_wcscasecmp
+*
+!*.c
+!*.cc
+!*.cpp
+!do_test_faccessat_setuid
--- a/test/test_fdopendir.c
+++ b/test/test_fdopendir.c
@@ -54,6 +54,7 @@ int main() {
     DIR *dir;
     char *first_entry = NULL;
     int err;
+    int pipefds[2];
 
     /* Test fdopendir with a valid directory fd, then use readdir */
 
@@ -205,6 +206,28 @@ int main() {
     /* Try to use fdopendir with AT_FDCWD - Should fail with EBADF */
 
     err = check_failure(AT_FDCWD, "AT_FDCWD", "EBADF", EBADF);
+    if (err) return 1;
+
+    /*
+     * Try to use fdopendir with pipe - Should fail with ENOTDIR
+     *
+     * This reproduces the stdin test failure seen when running the test via
+     * a pipe, rather than a tty or pty.
+     */
+
+    if (pipe(pipefds) < 0) {
+        perror("Unable to create pipe for pipe test");
+        return 1;
+    }
+    (void) close(pipefds[1]);  /* Close the writing side immediately */
+
+    err = check_failure(pipefds[0], "pipe", "ENOTDIR", ENOTDIR);
+    (void) close(pipefds[0]);
+    if (err) return 1;
+
+    /* Also try with closed pipe - Should fail with EBADF */
+
+    err = check_failure(pipefds[0], "closed", "EBADF", EBADF);
     if (err) return 1;
 
     return 0;
--- a/test/test_fdopendir.c
+++ b/test/test_fdopendir.c
@@ -26,8 +26,6 @@
 #include <stdio.h>
 #include <string.h>
 
-#define FEEDBACK 1
-
 /* Test expected failure case */
 
 int check_failure(int fd, const char *name, const char *exp_sym, int exp_val)
@@ -87,7 +85,7 @@ int main() {
             return 1;
         }
 
-#if FEEDBACK
+#ifdef FEEDBACK
         printf("%s\n", entry->d_name);
 #endif
     }
--- a/test/test_fdopendir.c
+++ b/test/test_fdopendir.c
@@ -28,12 +28,32 @@
 
 #define FEEDBACK 1
 
+/* Test expected failure case */
+
+int check_failure(int fd, const char *name, const char *exp_sym, int exp_val)
+{
+    DIR *dir;
+
+    if ((dir = fdopendir(fd))) {
+        fprintf(stderr, "error: fdopendir(%s) should have failed\n", name);
+        (void)closedir(dir);
+        return 1;
+    } else if (errno != exp_val) {
+    fprintf(stderr, "error: fdopendir(%s) failure should have returned"
+                    " %d (%s), actually returned %d (%s)\n",
+                    name, exp_val, exp_sym, errno, strerror(errno));
+        return 1;
+    }
+    return 0;
+}
+
 int main() {
     struct stat st;
     struct dirent *entry;
     int dfd = -1;
     DIR *dir;
     char *first_entry = NULL;
+    int err;
 
     /* Test fdopendir with a valid directory fd, then use readdir */
 
@@ -174,36 +194,18 @@ int main() {
 
     /* Try to use fdopendir with stdin - Should fail with ENOTDIR */
 
-    if ((dir = fdopendir(STDIN_FILENO))) {
-        fprintf(stderr, "error: fdopendir(stdin) should have failed\n");
-        (void)closedir(dir);
-        return 1;
-    } else if (errno != ENOTDIR) {
-        perror("error: fdopendir(stdin) should have failed with ENOTDIR");
-        return 1;
-    }
+    err = check_failure(STDIN_FILENO, "stdin", "ENOTDIR", ENOTDIR);
+    if (err) return 1;
 
     /* Try to use fdopendir with -1 - Should fail with EBADF */
 
-    if ((dir = fdopendir(-1))) {
-        fprintf(stderr, "error: fdopendir(-1) should have failed\n");
-        (void)closedir(dir);
-        return 1;
-    } else if (errno != EBADF) {
-        perror("error: fdopendir(-1) should have failed with EBADF");
-        return 1;
-    }
+    err = check_failure(-1, "-1", "EBADF", EBADF);
+    if (err) return 1;
 
     /* Try to use fdopendir with AT_FDCWD - Should fail with EBADF */
 
-    if ((dir = fdopendir(AT_FDCWD))) {
-        fprintf(stderr, "error: fdopendir(AT_FDCWD) should have failed\n");
-        (void)closedir(dir);
-        return 1;
-    } else if (errno != EBADF) {
-        perror("error: fdopendir(AT_FDCWD) should have failed with EBADF");
-        return 1;
-    }
+    err = check_failure(AT_FDCWD, "AT_FDCWD", "EBADF", EBADF);
+    if (err) return 1;
 
     return 0;
 }


fhgwright:
The current (preexisting) failures in the port context are:

10.6-10.8 `-universal` fail tests, due to some problem building `test_dirent_with_cplusplus` that doesn't occur when building "bare".

10.4 ppc `+universal` fails to build due to the lack of an adequate compiler.  Xcode gcc 4.0 is inadequate, and `apple-gcc42` can't currently build `+universal` targets on ppc.

11.x x86_64 fails to build `+universal`, because the code that knows not to use `optool` on arm64 only checks `build_arch`.

10.11-10.13 `+universal` fail `test_sysconf` with an abort trap.

tavianator (approved):

barracuda156:
@fhgwright What is the criterion of this working? I would like to test on 10.6 for `ppc`, since that has not been done, from the description (I can test on a standard 10.6.8 via Rosetta).

fhgwright:
@barracuda156 
> @fhgwright What is the criterion of this working? I would like to test on 10.6 for `ppc`, since that has not been done, from the description (I can test on a standard 10.6.8 via Rosetta).

As I said, I tested both "bare" (running directly in the repo) and via a locally-modified `legacy-support-devel` that points to this branch in my own fork.  I just pushed that to my `macports-ports` fork, so you can grab it from [https://github.com/fhgwright/macports-ports/tree/legacy-support-test](url) if you like.  Note that that branch is _not_ a PR, since it's pointing to an ephemeral commit, though a similar change can follow once this PR is merged.

No, it hasn't been tested on the extremely nonstandard 10.6 ppc. :-)

Since the substantive changes only affect the 10.4 build (where an error wouldn't have gone unnoticed) and failure cases of `fdopendir()`, it's probably not necessary to reset the clock on when this is release-ready.

Given the brokenness of the 1.1.x versions, it would probably make sense to call the next release 1.2.0.

fhgwright:
@barracuda156 
> @fhgwright What is the criterion of this working? I would like to test on 10.6 for `ppc`, since that has not been done, from the description (I can test on a standard 10.6.8 via Rosetta).

It would be an interesting data point to see if the 10.6 failure of `test_dirent_with_cplusplus` also occurs on ppc.  I have a suspicion that it relates to the `cxx_stdlib` stuff.

mascguy:
Fred, are you ready to have this merged, at least for a new release of `legacy-support-devel`?

barracuda156:
> @barracuda156
> 
> > @fhgwright What is the criterion of this working? I would like to test on 10.6 for `ppc`, since that has not been done, from the description (I can test on a standard 10.6.8 via Rosetta).
> 
> It would be an interesting data point to see if the 10.6 failure of `test_dirent_with_cplusplus` also occurs on ppc. I have a suspicion that it relates to the `cxx_stdlib` stuff.

@fhgwright I will run tests today.

fhgwright:
@mascguy 
> Fred, are you ready to have this merged, at least for a new release of `legacy-support-devel`?

Yes, even though @barracuda156 hasn't weighed in on 10.6 `ppc` yet, since it's a blocker for updating `legacy-support-devel`.

In the `Portfile`, I already have a fix locally for the 11.x `+universal` build problem, and expect to have a fix for the 10.6-10.8 test build failures shortly, so I'll include those fixes in the PR for the `legacy-support-devel` update.  I was correct in guessing that the 10.6-10.8 failures are related to the `cxx_stdlib` selection.  I haven't further investigated the 10.11-10.13 `+universal` test failures yet.

barracuda156:
@fhgwright Allow me an hour or so, I will return to this shortly.

fhgwright:
@barracuda156 
> @fhgwright Allow me an hour or so, I will return to this shortly.

OK, though I'm not the one doing the merging. :-)

I don't think you'll see the 10.6 test build failures on `ppc`, since they only happen when the MacPorts `cxx_stdlib` mismatches the OS default `cxx_stdlib`, which AFAIK currently depends on `clang`, which AFAIK isn't currently supported on `ppc`.  They happen when MacPorts stuffs `-stdlib=libc++` into `CXXFLAGS`, without a corresponding adjustment to the library selection.

barracuda156:
@fhgwright Test log from 10.6 ppc, from 07b4b1921508fb38ed302299ee94f90d6b0a3f30 commit:
[legacy-support-devel_07b4b1921508fb38ed302299ee94f90d6b0a3f30.txt](https://github.com/macports/macports-legacy-support/files/14337110/legacy-support-devel_07b4b1921508fb38ed302299ee94f90d6b0a3f30.txt)

fhgwright:
@barracuda156 
> @fhgwright Test log from 10.6 ppc, from [07b4b19](https://github.com/macports/macports-legacy-support/commit/07b4b1921508fb38ed302299ee94f90d6b0a3f30) commit: [legacy-support-devel_07b4b1921508fb38ed302299ee94f90d6b0a3f30.txt](https://github.com/macports/macports-legacy-support/files/14337110/legacy-support-devel_07b4b1921508fb38ed302299ee94f90d6b0a3f30.txt)

You failed to mentioned that this was a pass. :-) That is, except for some legacy cmath failures that are apparently not counted as real failures (and aren't anything new).  Your exit status was presumably zero.

@mascguy 
So we're good to go.

barracuda156:
@fhgwright Yes, all looks good. If it did not, I would have explicitly mentioned that something got broken :)

fhgwright:
I've now tracked down the 10.11-10.13 `+universal` test failures, which are actually due to a fairly bizarre reason.  It seems that Apple _never_ provided a working 32-bit version of `sysconf(_SC_PHYS_PAGES)`.  Prior to 10.11, that item isn't provided by the OS at all, so the working `legacy-support` version is used.  In 10.11-10.13 `+universal`, the test fails on the `i386` slice due to the broken native support.  In 10.14, `i386` is removed from `universal_archs`, so the `i386` case never occurs in a nominally `+universal` build, though it does fail as usual in a "bare" `i386` build.  In 10.15+, `i386` doesn't build.

This is a long-standing problem, so it's not a release blocker.

Assuming the `legacy-support` `sysconf` wrapper still works in 10.11-10.13, a simple fix would be to extend the use of the wrapper to all OS versions on `i386`.  That can wait until after the release.

fhgwright:
@fhgwright 
> Assuming the `legacy-support` `sysconf` wrapper still works in 10.11-10.13, a simple fix would be to extend the use of the wrapper to all OS versions on `i386`. That can wait until after the release.

I've verified that this approach works, but I'll hold off on submitting a PR until after the release, and leave this PR here as the release candidate.

macportsraf:
> Further use of dirfd outside the DIR functions is not allowed

Just a minor point about the quoted comment above. This sentence is not true. There was a manpage somewhere that made this same claim, and it was taken at face value, which lead to the initial problem of the `fdopendir()` implementation closing the fd immediately. The standard lists specific actions that are not allowed. It does not forbid all further uses. I think this sentence should be deleted. Or changed to list the specific actions that POSIX forbids (but only if it seems like a good idea to be documenting POSIX here). The whole purpose of this new fdopendir implementation is to allow futher use of dirfd outside the DIR functions (such as openat etc.).

fhgwright:
I just pushed that change.  I don't think it's necessary to rerun the tests on all platforms just for the deletion of one comment line, though I did do a quick run on my main system (10.9).

mascguy:
Fred/All, are we ready to merge?

fhgwright:
@mascguy 
> Fred/All, are we ready to merge?

Given that:

1. All my testing builds and passes except for the known issues noted.
2. @barracuda156 reports success on 10.6 `ppc`
3. I made the comment fix requested by @macportsraf 

I'd say yes.

I'll make a PR for the -devel port as soon as I can after this is merged.

mascguy:
> > Fred/All, are we ready to merge?
> 
> I'd say yes.
> 
> I'll make a PR for the -devel port as soon as I can after this is merged.

Perfect, merging.

mascguy (approved):

mascguy:
Oh, you have seven commits. Would you like to retain those as-is, squash a bit first, ...?

fhgwright:
@mascguy 
> Oh, you have seven commits. Would you like to retain those as-is, squash a bit first, ...?

No, I kept those separate as they're logically separate changes.  In particular, note that the fixes are in a separate commit from the new tests so that anyone wanting to reproduce the test failures can simply revert the fix commit while keeping the test commit.  And the test changes are split into one for refactoring (with no functional change) and one to add the new tests.

In general, I think "squashed and minimized" should mean "clean up whatever multiple commits you were using during development, but still keep multiple commits where appropriate".  And keep them ordered such that there are no intermediate broken states.

fhgwright:
Has this had a long enough soak period (9+ days) to justify a release?  Once there's a new release version, I'll make an updated Portfile and run it on my usual test stable (just for completeness, since it's the same code I tested before).

As I previously mentioned, this should probably be called 1.2.0.

mascguy:
> Has this had a long enough soak period (9+ days) to justify a release?  Once there's a new release version, I'll make an updated Portfile and run it on my usual test stable (just for completeness, since it's the same code I tested before).
> 
> As I previously mentioned, this should probably be called 1.2.0.

Fine with me, I'll create a new release.

mascguy:
> Fine with me, I'll create a new release.

Done:

https://github.com/macports/macports-legacy-support/releases/tag/v1.2.0

barracuda156:
@mascguy We would need to revbump a number of ports, I believe (as was discussed in the earlier PR).

mascguy:
> @mascguy We would need to revbump a number of ports, I believe (as was discussed in the earlier PR).

Good to know! So Fred would need to rev-bump those, when updating `legacy-support`.

barracuda156:
Yeah, right, this release is not the port yet.

fhgwright:
> > @mascguy We would need to revbump a number of ports, I believe (as was discussed in the earlier PR).
> 
> Good to know! So Fred would need to rev-bump those, when updating `legacy-support`.

Ah, yes.  I'd forgotten about that mess.  In a way, it's a crock to change the API incompatibly without changing the API version, but OTOH there may be only a relatively small subset of dependents that are impacted, while an API version change would require revbumping *all*  the dependents.  Investigation is needed.

fhgwright:
> > > @mascguy We would need to revbump a number of ports, I believe (as was discussed in the earlier PR).
> > 
> > 
> > Good to know! So Fred would need to rev-bump those, when updating `legacy-support`.
> 
> Ah, yes. I'd forgotten about that mess. In a way, it's a crock to change the API incompatibly without changing the API version, but OTOH there may be only a relatively small subset of dependents that are impacted, while an API version change would require revbumping _all_ the dependents. Investigation is needed.

@mascguy 
Out of paranoia, I just submitted a PR to make the two of us maintainers of the port, removing openmaintainer (and jonesc), and added a comment about not updating the port until the compatibility issue is sorted out.  I wish I'd remembered that issue _before_ calling for the release.
No other changes to it yet.

barracuda156:
@fhgwright I am afraid we do not have any way to know for sure which ports are affected. I have listed a few which must be revbumped, but those are merely which happened to be installed on my system. That is not a complete list be any means.

The only two approaches are either revbump everything which is potentially affected or revbump a few that we know should be revbumped and then hope someone notices breakages later.

fhgwright:
> @fhgwright I am afraid we do not have any way to know for sure which ports are affected. I have listed a few which must be revbumped, but those are merely which happened to be installed on my system. That is not a complete list be any means.
> 
> The only two approaches are either revbump everything which is potentially affected or revbump a few that we know should be revbumped and then hope someone notices breakages later.

"Everything potentially affected" means all dependents.  Based on grepping the indexes, the dependency counts (from a `selfupdate` yesterday) are:
```
darwin_8_i386:     1885
darwin_8_powerpc:  1885
darwin_9_i386:     1878
darwin_9_powerpc:  1878
darwin_10_i386:    1867
darwin_11_i386:    1722
darwin_12_i386:    1722
darwin_13_i386:    1712
darwin_14_i386:    1685
darwin_15_i386:    1662
darwin_16_i386:    1530
darwin_17_i386:      55
darwin_18_i386:      41
darwin_19_i386:      17
darwin_20_arm:        2
darwin_20_i386:       2
darwin_21_arm:        2
darwin_21_i386:       2
darwin_22_arm:        2
darwin_22_i386:       2
darwin_23_arm:        2
darwin_23_i386:       2
```
The darwin_20+ cases are presumably just ports (`rust-bootstrap-10.5` and `gutils-suite-base`) that failed to make the dependency conditional on the OS version.  I think Apple made a significant effort to bring macOS up to modern POSIX standards in 10.13, which is why there's such a huge drop at that point.  This specific issue shouldn't apply to darwin_14+, but that doesn't really help.

Clearly revbumping >1800 ports is unacceptable (Ken wasn't too far off when he said "thousands").

Just tossing it over the fence and seeing who screams is also unacceptable.

And I don't know of any practical way to actually check all the dependents to see which ones care.

Thus, the only reasonable approach I can see is to bring back the 8 `__mpls_` functions as transparent wrappers around the OS calls.  In theory, there's probably some way to arrange to pass the addresses of the OS calls through directly as the `__mpls_` addresses, but it would take some time to figure that out, and since most of these functions involve OS calls anyway, it's not worth worrying about a few extra machine instructions in the call.  And those would gradually fade away as the dependents get naturally updated.

fhgwright:
I just submitted PR #73 for the compatibiity fix.

barracuda156:
@fhgwright That will be a regression back to the semi-broken implementation.

What we need is a revbump, not bringing back what is known not to work correctly.

I am fine with revbumping a subset, there is no demand made to revbump everything.

fhgwright:
> @fhgwright That will be a regression back to the semi-broken implementation.

Nope.  See my comments in the other PR.

macportsraf:
We discussed which ports would need a revbump earlier. My thinking was that it would be all the ports that use legacysupport with `newest_darwin_requires_legacy` set to 13 or higher because they might be using `fdopendir`. That's approximately 159 ports. If the default value of `newest_darwin_requires_legacy` is 13 or higher, then any ports that don't specify it (160) would be included as well. I'm still happy to do a mass revbump pull request for the potentially affected ports. But it was decided that ports would only be revbumped when a problem was reported because the problems would only surface on old systems, so that approach would affect fewer people than a mass revbump of all potentially affected ports (approx 159 or 319).

It should be possible to reduce the number even further by scanning the source code of these ports looking for fdopendir. If it's not present, they don't need to be revbumped, but that would be a lot more work (and maybe impractical).

I don't understand how the thousands of ports number was arried at, so maybe I'm mistaken about something. But there are only 501 Portfiles that mention legacysupport, so I think that would have to be the upper bound. Presumably, I'm missing something.

macportsraf:
> That will be a regression back to the semi-broken implementation.

That PR is not the same as the semi-broken implementation because it only brings back the __mpls symbols but it doesn't bring back the old/recent legacysupport modified definition of DIR which was the cause of the problem that was eventually solved. A port (clang-devel) had failed to compile on 10.9 when using 10.10's SDK which has its own fdopendir with the standard definition of DIR. The function signatures were different and the compiler (or linker?) didn't like it.

I just looked at the new PR and was initially concerned that the clients it is designed to help would have been compiled using the modified definition of DIR, rather than the standard definition, and so those clients might still fail, but I don't think so. The clients only have a pointer to DIR so it shouldn't matter how DIR itself is defined (unless they do something like `sizeof(*dirp)` but that would be very wierd and very unlikely).

fhgwright:
This probably isn't the most appropriate PR to be discussing this, but...

@macportsraf 
> We discussed which ports would need a revbump earlier. My thinking was that it would be all the ports that use legacysupport with `newest_darwin_requires_legacy` set to 13 or higher because they might be using `fdopendir`. That's approximately 159 ports. If the default value of `newest_darwin_requires_legacy` is 13 or higher, then any ports that don't specify it (160) would be included as well. I'm still happy to do a mass revbump pull request for the potentially affected ports. But it was decided that ports would only be revbumped when a problem was reported because the problems would only surface on old systems, so that approach would affect fewer people than a mass revbump of all potentially affected ports (approx 159 or 319).
> 
> It should be possible to reduce the number even further by scanning the source code of these ports looking for fdopendir. If it's not present, they don't need to be revbumped, but that would be a lot more work (and maybe impractical).

All of the above is untrue.  See below.

> I don't understand how the thousands of ports number was arried at, so maybe I'm mistaken about something. But there are only 501 Portfiles that mention legacysupport, so I think that would have to be the upper bound. Presumably, I'm missing something.

That's based on grepping the indexes, rather than the Portfiles.  Since those contain actual processing of the Portfiles by `portindex`, it's a more reliable source than something which might not catch things done, e.g., via PortGroups.  Granted, the indexes treat each subport as a seprate port, but even allowing for that, the number is still huge.

> > That will be a regression back to the semi-broken implementation.
> 
> That PR is not the same as the semi-broken implementation because it only brings back the __mpls symbols but it doesn't bring back the old/recent legacysupport modified definition of DIR which was the cause of the problem that was eventually solved. A port (clang-devel) had failed to compile on 10.9 when using 10.10's SDK which has its own fdopendir with the standard definition of DIR. The function signatures were different and the compiler (or linker?) didn't like it.
> 
> I just looked at the new PR and was initially concerned that the clients it is designed to help would have been compiled using the modified definition of DIR, rather than the standard definition, and so those clients might still fail, but I don't think so. The clients only have a pointer to DIR so it shouldn't matter how DIR itself is defined (unless they do something like `sizeof(*dirp)` but that would be very wierd and very unlikely).

Given that DIR is dynamically allocated and freed by the directory functions, separately allocating it and/or embedding it in another structure is unlikely to be useful, as well as being rather difficult to exploit.  Given that the augmented DIR is larger than the normal DIR, any allocation based on the augmented DIR would contain the standard DIR just fine, without any risk of overflow.  That could be an issue when going *from* the standard DIR *to* the augmented DIR, but that ship has already sailed.  Switching in the current direction is fine.  And given that the added field in the augmented DIR is at the end, the offsets of the common elements are all unchanged, so references to them remain valid.  References to the added field should be gone in the new implementation.

Filtering based on `newest_darwin_requires_legacy` is neither significantly helpful nor safe.  The "not significantly helpful" part is because the number of dependents only drops from 1885 to 1712 when going from darwin_8 to darwin_13.  More importantly, the "unsafe" part is because the wrapped functions are the *traditional* directory functions which have been around forever, not fdopendir() and the "at" functions.  Any port built with `legacy-support` for OSX 10.9 or earlier that uses any of the 8 *traditional* directory functions is impacted.  Thus, locating all affected ports is a humongous undertaking, and still probably produces a humongous list of affected ports.

Furthermore, the revbump approach only works when the affected ports can actually be successfully rebuilt on all OS versions, which may often not be the case.  For 10.9 and earlier, the result of a failure is a port which is both outdated and broken.  For 10.10 and later, the outdated port isn't actually broken, but it still needs to be rebuilt to stop being "outdated", and until that time, every `port upgrade outdated` will stumble over it, perhaps even multiple times in the same invocation (yes, the code is really that dumb).  The latter part could be avoided by making the revbump conditional on the OS version, but that would be highly nonstandard and probably severely frowned upon.

Also consider that many MacPorts developers would like to see the older systems go away altogether, and doing something with a massive widespread impact, solely to help with an issue that only affects systems more than 10 years old, is likely to increase the hostility to supporting older systems, possibly even to the extent of making `legacy-support` "porta non grata".

Hence, the revbump approach is a non-starter.  Fortunately, it's now unnecessary.

macportsraf:
Thanks. I wasn't thinking about opendir itself etc. By the way, the augmented DIR was smaller than the normal one, not larger (it containing a DIR* and an int), but I really don't think it matters. I don't think any port would be doing something that would be affected.

fhgwright:
@macportsraf 
> Thanks. I wasn't thinking about opendir itself etc. By the way, the augmented DIR was smaller than the normal one, not larger (it contained a DIR* and an int), but I really don't think it matters. I don't think any port would be doing something that would be affected.

Ah, yes, you're right.  But indeed it's highly unlikely that any port would do that.

I believe it was always the intent for DIR to be regarded as opaque, which could have been enforced by using an incomplete struct definition in the public header.  But then the new `fdopendir()` would have had to do an end run around that in some fashion. :-)

PR #73, opened 2024-03-03
Add compatibility entries for obsolete __mpls_* funcs.

fhgwright:
The __mpls_* wrappers are no longer needed for the new fdopendir() implementation, but many existing dependents may still have references to them.  This adds those functions back as transparent wrappers for the corresponding OS calls.

This also adds tests for the wrappers, by applying the existing tests (in test_fdopendir) to the wrappers via macro replacements.

TESTED:
Tested on 10.4-10.5 ppc, 10.5-10.6 ppc (x86_64 Rosetta), 10.4-10.6 i386, 10.4-12.x x86_64, and 11.x-14.x arm64.
Builds on all tested platforms except 10.5 ppc +universal, and passes all tests in all buildable cases, including the new tests. Comparison with "nm" shows that no global definitions were removed from the libraries since v1.1.1, with additions as needed for the new fdopendir(), and no differences at all for OS 10.10+.

### DIFF
--- a/src/dirfuncs_compat.c
+++ b/src/dirfuncs_compat.c
@@ -1,0 +1,75 @@
+/*
+ * Copyright (c) 2024 Frederick H. G. Wright II <fw@fwright.net>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+/*
+ * Earlier versions of legacy-support needed wrappers around several OS
+ * calls related to directories, in order to implement fdopendir().  That
+ * is no longer the case, but existing dependents still reference those
+ * wrapper calls.  For compatibility, we continue to provide those functions,
+ * but just as transparent wrappers around the OS calls.
+ *
+ * This is only relevant for OS versions where our fdoopendir() is needed,
+ * hence the conditional (which is the same conditional as was used for
+ * the earlier implementations).
+ */
+
+/* MP support header */
+#include "MacportsLegacySupport.h"
+#if __MP_LEGACY_SUPPORT_FDOPENDIR__
+
+#include "dirfuncs_compat.h"
+
+DIR *
+__mpls_opendir(const char *filename) {
+    return opendir(filename);
+}
+
+struct dirent *
+__mpls_readdir(DIR *dirp) {
+    return readdir(dirp);
+}
+
+int
+__mpls_readdir_r(DIR *dirp, struct dirent *entry, struct dirent **result) {
+    return readdir_r(dirp, entry, result);
+}
+
+long
+__mpls_telldir(DIR *dirp) {
+    return telldir(dirp);
+}
+
+void
+__mpls_seekdir(DIR *dirp, long loc) {
+    seekdir(dirp, loc);
+}
+
+void
+__mpls_rewinddir(DIR *dirp) {
+    rewinddir(dirp);
+}
+
+int
+__mpls_closedir(DIR *dirp) {
+    return closedir(dirp);
+}
+
+int
+__mpls_dirfd(DIR *dirp) {
+    return dirfd(dirp);
+}
+
+#endif /* __MP_LEGACY_SUPPORT_FDOPENDIR__ */
--- a/src/dirfuncs_compat.h
+++ b/src/dirfuncs_compat.h
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) 2024 Frederick H. G. Wright II <fw@fwright.net>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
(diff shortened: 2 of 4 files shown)


barracuda156:
Please, do not merge this. This gonna bring back the disaster we had with a broken legacysupport.

fhgwright:
> Please, do not merge this. This gonna bring back the disaster we had with a broken legacysupport. (At least potentially: I cannot test anything at the moment, and related issues may not even be trivial to detect.)

WTF ARE YOU TALKING ABOUT???

The old implementation was broken because it relied on wrapping various OS functions in order to change their behavior to interface them to the old fdopendir() implementation.  This change simply adds TRANSPARENT wrappers for the OS calls so that existing dependents work as is, making exactly the same OS calls as they would get when rebuilt with the new headers (except for a few additional machine instructions in the wrappers).  In fact, Ken even alluded to doing this to avoid the compatibility problem, but didn't volunteer to do the work.

Did you even bother to read the change, which makes no changes to existing code other than expanding a comment, and then adds three new files for the wrappers and the tests?

> What we need is revbump of affected ports, not the return of something we know for sure not to work correctly.

It is NOT a "return of something we know for sure not to work correctly", since the latter is still gone.  RTFC.

> (Of course, if there are problems with existing implementation, those have to be addressed, but avoiding a need to revbump is not a reason to go for untested hacks. Neither it is a reason to add an extra layer of code.)

What we do NOT need is having to revbump an unknown massive number of ports, some of which may no longer even build on some platforms.

There are definitely still problems with the existing implementation, such as that the ATCALL mechanism isn't signal-safe, as well as the known thread-safety issue on 10.4, but none of that is new, and isn't a release blocker.

mascguy:
> The old implementation was broken because it relied on wrapping various OS functions in order to change their behavior to interface them to the old fdopendir() implementation.  This change simply adds TRANSPARENT wrappers for the OS calls so that existing dependents work as is, making exactly the same OS calls as they would get when rebuilt with the new headers (except for a few additional machine instructions in the wrappers).  In fact, Ken even alluded to doing this to avoid the compatibility problem, but didn't volunteer to do the work.

Makes sense. And if this eliminates the need for rev-bumps, even better.

Are there any specific ports/scenarios where the fix can be validated?

mascguy (approved):
LGTM, though haven't validated the fixes locally.

@kencu ?

fhgwright:
@mascguy 
> > The old implementation was broken because it relied on wrapping various OS functions in order to change their behavior to interface them to the old fdopendir() implementation.  This change simply adds TRANSPARENT wrappers for the OS calls so that existing dependents work as is, making exactly the same OS calls as they would get when rebuilt with the new headers (except for a few additional machine instructions in the wrappers).  In fact, Ken even alluded to doing this to avoid the compatibility problem, but didn't volunteer to do the work.
> 
> Makes sense. And if this eliminates the need for rev-bumps, even better.
> 
> Are there any specific ports/scenarios where the fix can be validated?

I didn't test it myself beyond running the existing internal tests (including the ones I modified to check the new entries).  But what I know is:

1. No previously existing code is modified by this change (though a comment is expanded).
2. The newly added entry points should not be referenced by anything building against the current headers.
3. Existing dependents not referencing those entry points are unaffected.
4. Existing dependents that do reference those entry points will fail to load before this change.
5. Existing dependents that reference those entry points will continue to work after this change, with the same updated functional behavior as they would have when rebuilt against the current headers (where they use the OS calls directly).

Or, as the saying goes, "all good things, no bad things".

Once this is merged, I can make a new `legacy-support-devel` pointing at it, which makes testing it with dependents easier.  Something I hadn't noticed until digging into dependents is that they all seem to use path dependencies, which means that simply deactivating `legacy-support` and activating `legacy-support-devel` is sufficient to cause all installed dependents to use the new code.  As far as this entire round of changes is concerned, it's only relevant on 10.9 and earlier.

To reiterate the dependent situation, the counts (without knowing which ones use these specific calls) look like:
```
darwin_8_i386:     1885
darwin_8_powerpc:  1885
darwin_9_i386:     1878
darwin_9_powerpc:  1878
darwin_10_i386:    1867
darwin_11_i386:    1722
darwin_12_i386:    1722
darwin_13_i386:    1712
darwin_14_i386:    1685
darwin_15_i386:    1662
darwin_16_i386:    1530
darwin_17_i386:      55
darwin_18_i386:      41
darwin_19_i386:      17
darwin_20_arm:        2
darwin_20_i386:       2
darwin_21_arm:        2
darwin_21_i386:       2
darwin_22_arm:        2
darwin_22_i386:       2
darwin_23_arm:        2
darwin_23_i386:       2
```
As I said in the other thread, I'm pretty sure the `darwin_20+` cases are bugs.

mascguy:
> Once this is merged, I can make a new `legacy-support-devel` pointing at it, which makes testing it with dependents easier.  Something I hadn't noticed until digging into dependents is that they all seem to use path dependencies, which means that simply deactivating `legacy-support` and activating `legacy-support-devel` is sufficient to cause all installed dependents to use the new code.  As far as this entire round of changes is concerned, it's only relevant on 10.9 and earlier.

Sounds good, let's merge and update `legacy-support-devel`.

fhgwright:
@mascguy 
> Sounds good, let's merge and update `legacy-support-devel`.

PR now at [https://github.com/macports/macports-ports/pull/22885](url).

kencu:
> LGTM, though haven't validated the fixes locally.
> 
> @kencu ?

I believe this should work out, indeed, to bridge the transition and eventually, as everything gets rebuilt, might after some period of time be removed.

fhgwright:
@kencu 
> > LGTM, though haven't validated the fixes locally.
> > @kencu ?
> 
> I believe this should work out, indeed, to bridge the transition and eventually, as everything gets rebuilt, might after some period of time be removed.

Indeed, though determining when "everything" has been rebuilt is challenging, so this will probably need to stick around for the foreseeable future.  I should add a comment to that effect, but don't want to rock the boat until after the release.

It should be noted that this provides upward compatibility but not downward compatibility.  I.e., dependents built against the old headers will work with the new library, but dependents built against the new headers may malfunction with the old library.  That could be avoided by keeping the 8 functions pointing at the wrappers for another release cycle, but I don't think it's worth it.

The wrappers may be even cheaper than I thought, though I haven't inspected the code to check.  If the compiler (in an optimized build) notices that the wrapper does nothing but call the wrapped function and return its result, and that both functions have the same signature, it may reduce the "wrapper" to a single jump instruction.

fhgwright:
@mascguy 
Is it time to cut a new release yet (presumably 1.2.1)?  The latest `legacy-support-devel` has only been out for 5 days, but the last change is fairly low-risk.

The recent history is:

20240303:
Added compatibility entry points.  No change to existing code.

20240220:
Fixed some failure cases to report correct errors.  Added tests.  Cleaned up 10.4-specific hack.  Also a couple of Portfile fixes not limited to the -devel case.

2024013:
Last change affecting non-error cases.

Once there's a release, I'll create and test a PR for the port update.  At last no revbumps needed. :-)

mascguy:
> Is it time to cut a new release yet (presumably 1.2.1)? The latest `legacy-support-devel` has only been out for 5 days, but the last change is fairly low-risk.

Ugh, sorry for the delayed response.

Agreed, it's time for a new release. I'll create 1.2.1 ASAP.

mascguy:
> > Is it time to cut a new release yet (presumably 1.2.1)? The latest `legacy-support-devel` has only been out for 5 days, but the last change is fairly low-risk.

> Agreed, it's time for a new release. I'll create 1.2.1 ASAP.

Done: 

https://github.com/macports/macports-legacy-support/releases/tag/v1.2.1

fhgwright:
> > > Is it time to cut a new release yet (presumably 1.2.1)? The latest `legacy-support-devel` has only been out for 5 days, but the last change is fairly low-risk.
> 
> > Agreed, it's time for a new release. I'll create 1.2.1 ASAP.
> 
> Done:
> 
> https://github.com/macports/macports-legacy-support/releases/tag/v1.2.1

Thanks.  The updated Portfile is ready except for testing, which I should be able to do tomorrow.

PR #75, opened 2024-03-17
sysconf(_SC_PHYS_PAGES): Always use wrapper on i386

fhgwright:
Apple's implementation of sysconf(_SC_PHYS_PAGES) never worked on i386.  It wasn't provided until 10.11, which requires x86_64, so the problem is never seen in ordinary native builds.  It is visible in i386 builds for 10.11+, including universal builds with i386.  Since MacPorts dropped i386 from the default universal_archs in 10.14, it's not normally seen in 10.14+.  It can be seen in the test failures in the 10.11-10.13 +universal cases.

The fix is simply to use the working legacy-support implementation in all i386 builds, regardless of OS version.

TESTED:
Tested on 10.4-10.5 ppc, 10.5-10.6 ppc (x86_64 Rosetta), 10.4-10.6 i386, 10.4-12.x x86_64, and 11.x-14.x arm64.
Builds on all tested platforms except 10.4 ppc +universal, and passes all tests in all buildable cases, including the previously failing 10.11-10.13 +universal tests.

### DIFF
--- a/include/MacportsLegacySupport.h
+++ b/include/MacportsLegacySupport.h
@@ -118,8 +118,9 @@
 /* sys/aio.h header needs adjustment to match newer SDKs */
 #define __MP_LEGACY_SUPPORT_SYSAIOTIGERFIX__  (__APPLE__ && __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1050)
 
-/*  sysconf() is missing some functions on some systems */
-#define __MP_LEGACY_SUPPORT_SYSCONF_WRAP__    (__APPLE__ && __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 101100)
+/*  sysconf() is missing some functions on some systems, and may misbehave on i386 */
+#define __MP_LEGACY_SUPPORT_SYSCONF_WRAP__    (__APPLE__ && (__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 101100 \
+                                                             || defined(__i386)))
 
 /* pthread_rwlock_initializer is not defined on Tiger */
 #define __MP_LEGACY_SUPPORT_PTHREAD_RWLOCK__  (__APPLE__ && __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1050)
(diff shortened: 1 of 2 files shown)


mascguy (approved):
LGTM.

@kencu and anyone else...?

kencu:
I have not noticed or verified that the os-provided i386 version dosen’t work. Nor can I verify that this fixes it, and is used over the system version. Nor have I looked to see what exactly happens in a universal build, where presumably this now would now work different ways on the same system depending on arch….

Given all that, Fred owns this one and as maintainer now can fix any issues that show up.

mascguy:
> I have not noticed or verified that the os-provided i386 version dosen’t work. Nor can I verify that this fixes it, and is used over the system version. Nor have I looked to see what exactly happens in a universal build, where presumably this would now work different ways on the same system depending on arch….
> 
> Given all that, Fred owns this one and as maintainer now can fix any issues that show up.

Sounds good. I'll go ahead and merge Fred's two PRs, and create a new release of `legacy-support-devel`.

fhgwright:
@kencu
> I have not noticed or verified that the os-provided i386 version dosen’t work. Nor can I verify that this fixes it, and is used over the system version. Nor have I looked to see what exactly happens in a universal build, where presumably this would now work different ways on the same system depending on arch….

I think I mentioned this bug somewhere along the line in relation to the testing I did for PR 69.

It's easy to reproduce if you have a 10.11-10.13 system available - just do a `port test legacy-support +universal`.  The i386 slice fails the tests for this reason.  It's also possible to see it on 10.14, if you override the default `universal_archs` to include i386, which actually works in spite of beliefs to the contrary. On 10.15+, i386 builds fail, so the issue is moot.

Note that I didn't have to add a test for this bug, since the test was already present and failing.

As with all architecture-related tests done via the preprocessor, this does the right thing in universal builds.  Whether dependencies are handled correctly needs to be further investigated, though that's a Portfile issue (perhaps a PortGroup issue) not covered here.  In any case, this change won't make anything worse.  Existing affected dependents may or may not pick up the change automatically, depending on whether they're already using the `sysconf` wrapper for other reasons.

@mascguy 
Looks like the third time's a charm. :-)

Sometimes it seems like compiler authors go out of their way to maximize the amount of extra stuff one needs to avoid pointless warnings.  The code would have been perfectly legal without the `defined()`, since undefined preprocessor macros get treated as 0 in numeric preprocessor tests, but relying on that often provokes a _different_ warning.

Thanks for cleaning it up.

PR #84, opened 2024-04-21
test_dirfuncs_compat: Fix warning

fhgwright:
Since dirfd() may be defined as a macro, it needs to be undefined before defining it, to avoid a warning.

TESTED:
Tested on 10.4-10.5 ppc, 10.4-10.6 i386, 10.5-10.6 ppc (i386 Rosetta), 10.5-12.x x86_64, 11.x-14.x arm64.
Redfinition warning is now gone.

### DIFF
--- a/test/test_dirfuncs_compat.c
+++ b/test/test_dirfuncs_compat.c
@@ -35,6 +35,7 @@
 #define seekdir     __mpls_seekdir
 #define rewinddir   __mpls_rewinddir
 #define closedir    __mpls_closedir
+#undef dirfd
 #define dirfd       __mpls_dirfd
 
 #include "test_fdopendir.c"


fhgwright:
This isn't a high priority, so if a release is imminent it could be postponed to avoid rocking the boat.  OTOH, since it's a low-risk change that only affects a test, it might be reasonable not to worry about the usual `legacy-support-devel` cycle.

I should have caught this earlier, but the trouble with verbose build procedures is that warnings tend to get buried in all the chatter.  This isn't the only warning, but it's recently introduced and easily fixed.

mascguy:
> This isn't a high priority, so if a release is imminent it could be postponed to avoid rocking the boat. OTOH, since it's a low-risk change that only affects a test, it might be reasonable not to worry about the usual `legacy-support-devel` cycle.

I'm fine with either. But given the low-risk nature of this - and warning reductions are always good - I'd say let's go ahead and merge.

Great stuff, thanks as always Fred!

mascguy (approved):

fhgwright:
@macsguy 
> I'm fine with either. But given the low-risk nature of this - and warning reductions are always good - I'd say let's go ahead and merge.

I see you did that and also updated `legacy-support-devel` to match.

Given that there are no other updates on the short-term horizon, and my availability is likely to be reduced for the next 3 weeks, it might be a good idea to cut a new release now so that I can run a new Portfile through the usual testing.  In this case, it's not just the version bump, but also ripping out all the reexport workaround stuff.  I have a preliminary Portfile ready to go for v1.2.3.

mascguy:
> Given that there are no other updates on the short-term horizon, and my availability is likely to be reduced for the next 3 weeks, it might be a good idea to cut a new release now so that I can run a new Portfile through the usual testing.  In this case, it's not just the version bump, but also ripping out all the reexport workaround stuff.  I have a preliminary Portfile ready to go for v1.2.3.

Sounds good to me!

fhgwright:
@mascguy 
> > Given that there are no other updates on the short-term horizon, and my availability is likely to be reduced for the next 3 weeks, it might be a good idea to cut a new release now so that I can run a new Portfile through the usual testing.  In this case, it's not just the version bump, but also ripping out all the reexport workaround stuff.  I have a preliminary Portfile ready to go for v1.2.3.
> 
> Sounds good to me!

You'll need to cut the new release, since I can't.  That's a prerequisite for testing the updated Portfile.

mascguy:
> You'll need to cut the new release, since I can't. That's a prerequisite for testing the updated Portfile.

Done: https://github.com/macports/macports-legacy-support/releases/tag/v1.2.3

PR #85, opened 2024-05-02
stpncpy: Fix potential header collision with "later" SDK

fhgwright:
This is one case of the general problem in:

https://trac.macports.org/ticket/69867

This should fix the only known case specifically related to collisions with "secure" wrappers.  Whether there are other types of SDK collisions is TBD.

TESTED:
Builds and passes tests on 10.4-10.5 ppc, 10.4-10.6 i386, 10.5-10.6 ppc (i386 Rosetta), 10.5-12.x x86_64, 11.x-14.x arm64. Builds test_stpncpy successfully on 10.6 with 10.7 SDK.

### DIFF
--- a/include/string.h
+++ b/include/string.h
@@ -27,6 +27,7 @@
 /* stpncpy */
 #if __MP_LEGACY_SUPPORT_STPNCPY__
 __MP__BEGIN_DECLS
+#undef stpncpy  /* In case built with later SDK */
 extern char *stpncpy(char *dst, const char *src, size_t n);
 __MP__END_DECLS
 #endif


fhgwright:
> adding the single missed entry for the new stpncpy function to <secure/_string.h> would seem more robust to me, and will eventually be the fix.

Sigh.  How many times to I have to explain that adding "secure" support for `stpncpy` (at least in the normal way) would _not_ fix the problem?  The fix here would be needed _whether or not_ "secure" support for `stpncpy` is added.

And how many times do I have to explain that `legacy-support` does _not_ currently have _any_ "secure" support whatsoever?  None.  Zero.  Nada.  Adding such support, while desirable, wouldn't be limited to `stpncpy`, and again, wouldn't fix the problem at hand.

> However, you can have it like this for now if you want it like this.
> 
> Please just first confirm you tested on 10.6 against the 10.7 SDK with some _USE_FORTIFY_LEVEL set to greater than 0 to show this worked.

There's no need to set `_USE_FORTIFY_LEVEL` explicitly - it happens automatically.  As mentioned (albeit more tersely) in the commit summary, I tested it by compiling `test_stpncpy` on 10.6 with the 10.7 SDK, both with and without the fix.  Without the fix, it gets the same error as `R-nloptr`, and with the fix it compiles successfully.

fhgwright:
@mascguy 
As usual, merging this is a prerequisite for updating the -devel port.

fhgwright:
@kencu 
Just to clarify:

Everything related to secure wrappers necessarily happens _after_ the function is declared.

The `#undef` that fixes the conflict is necessarily required _before_ the function is declared.

Hence, the two are independent.  Any secure wrapper support will be _in addition to_, not _instead of_, the fix here.

I expect to have diminishing available time until after next week, so I'd like to get this put to bed while I still have time for testing.

kencu:
Fred -- your understanding of this issue -- sigh -- does not match mine -- sigh -- and I guess  -- sigh -- you're on your own -- sigh.

Sigh.

What a problem.

Sigh.

kencu:
Sigh-ning off.

fhgwright:
@mascguy 

> > Fred -- your understanding of this issue -- sigh -- does not match mine...
> 
> Okay, so was this ever documented anywhere... even if only in old technical manuals that have been out-of-print for 20+ years?
> 
> Or is the "documentation" limited to the source code for 20+ year-old toolchain components?
> 
> Or... what?
> 
> Regardless, if we can't come to a consensus, then we need to find a definitive source of truth.
> 
> Where do we go from here?

It has nothing to do with toolchains, and the only relevant source code is the header files themselves.  Since this seems to be surprisingly nonobvious, here's a detailed explanation:

For any given function with no "secure wrapper", the flow looks like this:

First we have the OS header, which we'll call A:

1) Declare the function, if included in this SDK.

Then we have the `legacy-support` header, which we'll call B:

1) Include the OS header (via include_next).
2) Declare the function, if missing in the target OS version.

When combined, we have C:

1.1) Declare the function, if included in this SDK.
2) Declare the function, if missing in the target OS version.

When the SDK version matches the target OS version, either step 1 or step 2 happens, but not both.  When the SDK version is later than the target OS version, both steps may happen.  This is OK, since duplicate declarations are allowed, as long as they are compatible.  If they _weren't_ compatible, that would be a serious bug, since the `legacy-support` and OS implementations of the same function damn well better have the same signature.

In the case where the function includes an optional "secure wrapper", A is replaced by D:

1) Declare the function, if included in this SDK.
2) Optionally, define the function (if included in this SDK) as a wrapper macro, which adds additional checks to each call.  This is done via an additional header, `#include`d at the bottom of the standard header.

The order of these steps is important, because once the function is defined as the wrapper macro, referencing it in a declaration is syntactically illegal and results in errors.  That's why the `#include` of the "secure" augmentation header happens at the _bottom_ of the main header, rather than at the top like the others.

Since `legacy-support` has no "secure wrapper" mechanism, its header remains B, which when combined with D yields E:

1.1) Declare the function, if included in this SDK.
1.2) Optionally, define the function (if included in this SDK) as a wrapper macro, which adds additional checks to each call.  This is done via an additional header, `#include`d at the bottom of the standard header.
2) Declare the function, if missing in the target OS version.

When the SDK version matches the target OS version, then either steps 1.1 and 1.2 happen, or step 2 happens, but not both.  When the SDK version is later than the OS version, all three steps may happen.  If they do, it results in syntax errors from step 2 due to the preceding step 1.2, as mentioned above in "order is important".

If "secure" support were added to the `legacy-support` header, then B would become F:

1) Include the OS header (via include_next).
2) Declare the function, if missing in the target OS version.
3) Optionally, if the function is missing in the target OS version, define it as a wrapper macro, adding the additional checks.  This could be done either directly in F, or in a separate header `#include`d at the bottom, paralleling the organization of the OS headers.

That would cause the combined version of D and F to become G:

1.1) Declare the function, if included in this SDK.
1.2) Optionally, define the function (if included in this SDK) as a wrapper macro, which adds additional checks to each call.  This is done via an additional header, `#include`d at the bottom of the standard header.
2) Declare the function, if missing in the target OS version.
3) Optionally, if the function is missing in the target OS version, define it as a wrapper macro, adding the additional checks.

When the SDK version matches the target OS version, then either steps 1.1 and 1.2 happen, or steps 2 and 3 happen, but not both.  When the SDK version is later than the OS version, all four steps may happen.  If they do, it results in syntax errors from step 2 due to the preceding step 1.2, as mentioned above in "order is important".

To fix the SDK mismatch issue (alone), we need to replace B with H:

1) Include the OS header (via include_next).
2) `#undef` the function, if missing in the target OS version.
3) Declare the function, if missing in the target OS version.

Combining D with H, we get J:

1.1) Declare the function, if included in this SDK.
1.2) Optionally, define the function (if included in this SDK) as a wrapper macro, which adds additional checks to each call.  This is done via an additional header, `#include`d at the bottom of the standard header.
2) `#undef` the function, if missing in the target OS version.
3) Declare the function, if missing in the target OS version.

When the SDK version matches the target OS version, then either steps 1.1 and 1.2 happen, or steps 2 and 3 happen, but not both.  If steps 2 and 3 (only) happen, step 2 is a NOP, since no previous macro definition exists.  When the SDK version is later than the OS version, all four steps may happen.  In this case, step 2 undoes step 1.2, making step 3 legal.

It's also possible to incorporate both the SDK compatibility fix and the secure wrapper enhancement into B, as K:

1) Include the OS header (via include_next).
2) `#undef` the function, if missing in the target OS version.
3) Declare the function, if missing in the target OS version.
4) Optionally, if the function is missing in the target OS version, define it as a wrapper macro, adding the additional checks.

Combining D with K, we get L:

1.1) Declare the function, if included in this SDK.
1.2) Optionally, define the function (if included in this SDK) as a wrapper macro, which adds additional checks to each call.  This is done via an additional header, `#include`d at the bottom of the standard header.
2) `#undef` the function, if missing in the target OS version.
3) Declare the function, if missing in the target OS version.
4) Optionally, if the function is missing in the target OS version, define it as a wrapper macro, adding the additional checks.

When the SDK version matches the target OS version, then either steps 1.1 and 1.2 happen, or steps 2-4 happen, but not both.  If steps 2-4 (only) happen, step 2 is a NOP, since no previous macro definition exists.  When the SDK version is later than the OS version, all five steps may happen.  In this case, step 2 undoes step 1.2, making step 3 legal, and step 4 happens as usual, reintroducing the wrapper.


To summarize:

```
 legacy-support header    Fixes SDK Issue   Adds secure wrapper
-----------------------   ---------------   -------------------
B                         No                No
F                         No                Yes
H                         Yes               No
K                         Yes               Yes
```

Hence, the two changes are completely unrelated, and apply to different issues.  Either or both can be added independently.  They don't even touch the same source lines, and if they were separate git branches, there would be a conflict-free merge between them.  And since they're unrelated changes for unrelated purposes, they should be separate git commits, anyway.

The change in this PR is the one-line addition for B->H.  Its obvious correctness has been validated by applying it to a repro of the original problem.

If the intent is to make the perfect the enemy of the good by holding the build-failure fix hostage to implementing the secure-wrapper mechanism, then either someone else will have to do it or it will have to wait until at least next week, since I have no time this week to implement and test the secure-wrapper change.

mascguy:
> To summarize:
> 
> ```
>  legacy-support header    Fixes SDK Issue   Adds secure wrapper
> -----------------------   ---------------   -------------------
> B                         No                No
> F                         No                Yes
> H                         Yes               No
> K                         Yes               Yes
> ```
> 
> Hence, the two changes are completely unrelated, and apply to different issues. Either or both can be added independently. They don't even touch the same source lines, and if they were separate git branches, there would be a conflict-free merge between them. And since they're unrelated changes for unrelated purposes, they should be separate git commits, anyway.
>
> The change in this PR is the one-line addition for B->H. Its obvious correctness has been validated by applying it to a repro of the original problem.

Makes sense, and thank you for the thorough explanation!

> If the intent is to make the perfect the enemy of the good by holding the build-failure fix hostage to implementing the secure-wrapper mechanism, then either someone else will have to do it or it will have to wait until at least next week, since I have no time this week to implement and test the secure-wrapper change.

I definitely don't want to slow things down, simply trying to ensure we take into account others' concerns too. But given your understanding of this topic, no worries here.

mascguy (approved):

PR #86, opened 2024-05-06
provide workaround for pthread_get_stacksize_np

MarcusCalhoun-Lopez:
see https://github.com/rust-lang/rust/issues/43347
fixes https://trac.macports.org/ticket/68015

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,11 @@ Wrapped headers and replaced functions are:
     <td>Adds <code>_fstatat$INODE64</code> library symbol</td>
     <td>OSX10.9</td>
   </tr>
+  <tr>
+    <td><code>-</code></td>
+    <td>Provides a workaround for bug in <code>pthread_get_stacksize_np</code></td>
+    <td>OSX10.4, OSX10.5, OSX10.9, OSX10.10</td>
+  </tr>
 </table>
 
 For information on building this library outside MacPorts, see BUILDING.txt.
--- a/include/MacportsLegacySupport.h
+++ b/include/MacportsLegacySupport.h
@@ -201,4 +201,11 @@
 /* library symbol _dirfd */
 #define __MP_LEGACY_SUPPORT_SYMBOL__dirfd__  (__APPLE__ && __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1080)
 
+/* fix bug in pthread_get_stacksize_np */
+/* see https://github.com/rust-lang/rust/issues/43347 */
+#define __MP_LEGACY_SUPPORT_PTHREAD_GET_STACKSIZE_NP_FIX__ (__APPLE__ && ( \
+                                                            __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ == 101000 || \
+                                                            __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ == 1090 || \
+                                                            __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ <  1060 ))
+
 #endif /* _MACPORTS_LEGACYSUPPORTDEFS_H_ */
--- a/src/add_symbols.c
+++ b/src/add_symbols.c
@@ -81,3 +81,10 @@ extern const char fstatat$INODE64_tmp7 __asm("$ld$add$os10.7$_fstatat$INODE64");
 extern const char fstatat$INODE64_tmp8 __asm("$ld$add$os10.8$_fstatat$INODE64"); __attribute__((visibility("default"))) const char fstatat$INODE64_tmp8 = 0;
 extern const char fstatat$INODE64_tmp9 __asm("$ld$add$os10.9$_fstatat$INODE64"); __attribute__((visibility("default"))) const char fstatat$INODE64_tmp9 = 0;
 #endif
+
+#if !(__MP_LEGACY_SUPPORT_PTHREAD_GET_STACKSIZE_NP_FIX__)
+extern const char pthread_get_stacksize_np_tmp4 __asm("$ld$add$os10.4$_pthread_get_stacksize_np"); __attribute__((visibility("default"))) const char pthread_get_stacksize_np_tmp4 = 0;
+extern const char pthread_get_stacksize_np_tmp5 __asm("$ld$add$os10.5$_pthread_get_stacksize_np"); __attribute__((visibility("default"))) const char pthread_get_stacksize_np_tmp5 = 0;
+extern const char pthread_get_stacksize_np_tmp9 __asm("$ld$add$os10.9$_pthread_get_stacksize_np"); __attribute__((visibility("default"))) const char pthread_get_stacksize_np_tmp9 = 0;
+extern const char pthread_get_stacksize_np_tmp10 __asm("$ld$add$os10.10$_pthread_get_stacksize_np"); __attribute__((visibility("default"))) const char pthread_get_stacksize_np_tmp10 = 0;
+#endif
--- a/src/pthread_get_stacksize_np.c
+++ b/src/pthread_get_stacksize_np.c
@@ -1,0 +1,70 @@
+/* MP support header */
+#include "MacportsLegacySupport.h"
+#if __MP_LEGACY_SUPPORT_PTHREAD_GET_STACKSIZE_NP_FIX__
+
+#include <pthread.h>
+#include <sys/resource.h>
+#include <dlfcn.h>
+#include <stdlib.h>
+
+#if __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ >= 1090
+/* private system call available on OS X Mavericks (version 10.9) and later */
+/* see https://github.com/apple-oss-distributions/libpthread/blob/ba8e1488a0e6848b710c5daad2e226f66cfed656/private/pthread/private.h#L34 */
+pthread_t pthread_main_thread_np(void);
+#endif
+
+#define kMaxThreadStackSize 0x40000000 /* from LLVM: 1 << 30 or 1Gb */
+
+size_t pthread_get_stacksize_np(pthread_t t) {
+#if __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ >= 1090
+     int is_main_thread = pthread_equal(t, pthread_main_thread_np());
+#else
+     /* taken from Apple Libc */
+     /* see https://github.com/apple-oss-distributions/Libc/blob/224a8105d573730ae33f474ae5b63b113123aee4/pthreads/pthread.c#L1167 */
+     /* for _PTHREAD_CREATE_PARENT, see https://github.com/apple-oss-distributions/Libc/blob/224a8105d573730ae33f474ae5b63b113123aee4/pthreads/pthread_internals.h#L647C9-L647C31 */
+     /* for pthread_t, see https://github.com/apple-oss-distributions/Libc/blob/224a8105d573730ae33f474ae5b63b113123aee4/pthreads/pthread_internals.h#L107 */
+     /* for pthread_lock_t, see https://github.com/apple-oss-distributions/Libc/blob/224a8105d573730ae33f474ae5b63b113123aee4/pthreads/pthread_machdep.h#L214C13-L214C27 */
+     struct
+     {
+          long sig;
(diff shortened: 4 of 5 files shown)


mascguy (approved):
Based on a cursory review of the upstream ticket, etc, this looks good to me.

And since this may fix Rust build issues on 10.9/10.10, it would be good to merge sooner rather than later.

Anyone else... @fhgwright, @kencu, others... thoughts? (Even if you folks can only spare 5 minutes to review, that would still be great.)

fhgwright:
I don't have time to analyze this in detail, though superficially it looks plausible.  The one thing I do note is that the fix and the test are in the same commit, but it's better for testing if they're separate, so that one can demonstrate the problem with the test and no fix, and then demonstrate the fix by including both.  During development, it's often useful to create and commit the test first, but the published version should put the fix first, so that it doesn't create a pothole in the history for future bisects to stumble over.

If it were split in this fashion, I _might_ have time to run both states through my VMs, though not necessarily today.  This week is busy.

It sounds like `rust` might not make it to a full year of brokenness on 10.9. :-)

fhgwright:
> As requested, I split the test off into a separate commit.

OK, I ran both the test-only and test+fix cases on:
```
macOS 12.7.4 21H1123, x86_64, Xcode 14.2 14C18
macOS 12.7.4 21H1123, arm64, Xcode 14.2 14C18
macOS 13.6.6 22G630, arm64, Xcode 15.2 15C500b
macOS 14.4.1 23E224, arm64, Xcode 15.3 15E204a
```
The good news is that, on ppc, i386, and x86_64, the test-only case fails on 10.4, 10.5, 10.9 and 10.10, while passing on the others (presumably all as expected), and the test+fix case passes on all of them.

The bad news is that the test fails in all arm64 cases, with or without the fix.  That may simply be that the test is invalid on arm64, so it needs to either suppress that case or handle it properly.  If there's any possibility that this issue could arise on arm64, then testing that would be useful but not immediately mandatory (if it's not currently causing trouble).  The fact that the fix makes no difference on arm64 isn't surprising, given that it appears to do nothing on 11.x+ systems.

> If and when everyone is ok with this PR, we can squash the two commits.

IMO, they should be kept as separate commits permanently, except for swapping the order so that the fix comes first and there's no intermediate broken state in the history.  That way, if anyone wants to reproduce the issue in the future, they can simply revert the fix commit without reverting the test commit, with no need to resort to path checkouts. The test before fix order is more convenient during development, though.  The test commit should get a better description, of course. :-)

> I labeled this PR as "invalid" to try to help us remember to squash the commits.

The usual way to do that is to mark it as a draft.  The "remember" aspect is also relevant to reordering, of course.

MarcusCalhoun-Lopez:
> The bad news is that the test fails in all arm64 cases, with or without the fix.

Thank you for testing this PR.
I have updated the PR to hopefully properly handle Apple silicon machines.
 
> > If and when everyone is ok with this PR, we can squash the two commits.
> 
> IMO, they should be kept as separate commits permanently, except for swapping the order so that the fix comes first and there's no intermediate broken state in the history.

I have no strong feelings about this.
Both commit messages should now be valid, so switching the order should be easy once all the tests pass.

> 
> > I labeled this PR as "invalid" to try to help us remember to squash the commits.
> 
> The usual way to do that is to mark it as a draft. The "remember" aspect is also relevant to reordering, of course.

Thank you for pointing this out.
For clarity, this PR is now marked as a draft.

fhgwright:
@MarcusCalhoun-Lopez 
> I have updated the PR to hopefully properly handle Apple silicon machines.

The good news is that the arm64 failures are now gone.

The bad news is that there are now more failures, including with the fix.  This is the same set of test systems as before.

Without the fix, there are now failures on:
```
10.4-10.5 ppc
10.4-10.5 i386
10.5 x86_64
10.9-10.12 x86_64 (10.11-10.12 new)
```
With the fix, there are failures on:
```
10.5 ppc (real ppc only, not Rosetta)
10.5 i386
10.5 x86_64
10.12-10.14 x86_64 (this is bizarre)
```
It's possible that I did something wrong in the testing, but I don't have any more time to spend on it this week.  Next week I can check more carefully, and/or test another update, and/or provide logs of cases you can't reproduce.

It should be trivial to fix line 30 of the test, though:
```
     printf("Stack size for detached detached non-main thread seems to be ok\n");
```

MarcusCalhoun-Lopez:
I tried to simplify the test to **only** check if the stack size was larger on the main thread.
I also tried to ensure the threads did not exit too soon by using `sleep`.
Hopefully, this more accurately tests the new `pthread_get_stacksize_np`.

fhgwright:
@MarcusCalhoun-Lopez 
> I tried to simplify the test to **only** check if the stack size was larger on the main thread. I also tried to ensure the threads did not exit too soon by using `sleep`. Hopefully, this more accurately tests the new `pthread_get_stacksize_np`.

I'll retest when I get a chance, though as I said, I won't have time before next week.

I wasn't _too_ surprised to see new failures given that you'd added new tests, but seeing the fix make some cases go from passing to failing on OS versions where it wasn't supposed to be doing anything at all was pretty bizarre.

MarcusCalhoun-Lopez:
> I wasn't _too_ surprised to see new failures given that you'd added new tests, but seeing the fix make some cases go from passing to failing on OS versions where it wasn't supposed to be doing anything at all was pretty bizarre.

The fix for 10.5 and 10.4 makes an educated guess about the undocumented layout of `pthread_t` in memory.
Specifically, that there is and 8-bit field that is called `detached` in the source code.
For that reason, I thought it prudent to test both detached and joinable threads.
I *think* the reason for the new failures was that the detached threads were finishing before `pthread_get_stacksize_np` was called.
I added `sleep` calls in an attempt to prevent this.

fhgwright:
I've now had a chance to retest with the v3 test (same configurations as before).  Failures without the fix are now _almost_ (see below) limited to the expected 10.4-10.5 and 10.9-10.10 cases, and all the expected failures are eliminated by the fix.

The "almost" is that there's now a failure on 10.6 ppc, which isn't an expected failure case, and (unsurprisingly) isn't fixed by the fix.  That's not an officially supported configuration, though `legacy-support` has been working on it so far, so if there's a simple fix for that case, we might as well do it.  Although I haven't been masochistic enough so far to wrestle with a native 10.6 ppc install, I do have a 10.6 i386 VM with `build_arch` set to `ppc`, so it targets Rosetta.

The 10.6 ppc failure message is:
```
Assertion failed: (stack_size_thread==stack_size_detached && "Unknown Error, non-main threads are assumed to have the same stack size"), function main, file test/test_pthread_get_stacksize_np.c, line 33.
```

As far as the "sleep fix" goes, it seems to work but isn't really an ideal fix for the race, for a couple of reasons:

1. If the test is run on a sufficiently busy system, the main thread could get delayed longer than the sleep duration, causing the race to reappear.
2. The sleep unconditionally adds 5 seconds to the test runtime, which isn't huge but it does more than double the total duration of all tests, even without any parallelism.

I think a more robust and less time-wasting approach would be to have a flag that the main thread sets to give the subthread permission to exit.  The loop testing the flag would need a *short* sleep (e.g. `usleep(1000)`)to avoid a scenario (mainly on a single-core machine) where the main thread is starved waiting for the subthread to run out its quantum uselessly testing the flag that the main thread isn't getting a chance to set.

If you need more info from the 10.6 ppc case, let me know.

fhgwright:
@MarcusCalhoun-Lopez 
> Thank you for testing this. My multithreading skills are quite rusty. I modified the test to avoid any race conditions (I hope) and use `usleep(1000)`.

I spotted a bug by inspection after launching the VMs but before firing up the other machines, so I only ran the x86 VM cases this time around, though it looks good in those cases (including 10.6 ppc via Rosetta).

The bug is that, in this approach, it needs to wait for _both_ stack sizes to be reported before proceeding, since it can't assume that the two threads do that in the order that they were launched.

I notice a philosophical change between the v3 and v4 tests, where v3 had the main thread obtaining the stack sizes of the subthreads (with a bug), while v4 has each thread reporting its own stack size.  I presume either approach is valid, but don't know enough about the details to know whether it's worth doing it both ways for extra thoroughness, or whether that would be pointless.

The bug in v3 is that the code to get the detached thread's stack size was referencing the non-detached thread instead.  I don't know whether that was the cause of the 10.6 ppc failure, though it's probably not worth chasing.

So as a minimum, the `while` condition needs to be expanded.  I'll leave it to you as to whether checking the stack sizes both ways is worth the trouble.  If it is, then the general idea is that each subthread would have its own similar `while` loop to delay exiting until the main thread has read its stack size.

Minor style quibble: the "==" in line 46 should have blanks around it.

> > If you need more info from the 10.6 ppc case, let me know.

Which may no longer be necessary at this point.

> Yes, thank you, I will need help on 10.6 ppc. The best I can do is test on an 10.6 virtual machine with the standard macOS installed. `port test legacy-support build_arch=ppc` works for me.

That's essentially what I'm doing here, except that I'm just testing the project code directly as copied from the repo, rather than through the whole port mechanism.  I might _also_ do the latter as a final test after everything otherwise looks good, but there's no reason to iterate the extra hassle.

MarcusCalhoun-Lopez:
> So as a minimum, the `while` condition needs to be expanded. I'll leave it to you as to whether checking the stack sizes both ways is worth the trouble. If it is, then the general idea is that each subthread would have its own similar `while` loop to delay exiting until the main thread has read its stack size.

It is likely that checking both joinable and detached threads is overkill.
However, since the fix for 10.4 and 10.5 makes use of the 8-bit field called `detached`, it might be worth the trouble.

I would respectfully submit that the `while` loop is **only** necessary for the detached thread.
The joinable thread has [pthread_join](https://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_join.html), which "suspend[s] execution of the [main] thread until the [joinable] thread terminates," which is all the `while` loop is doing.

> Minor style quibble: the "==" in line 46 should have blanks around it.

Should be fixed.

fhgwright:
@MarcusCalhoun-Lopez 
> I would respectfully submit that the `while` loop is **only** necessary for the detached thread. The joinable thread has [pthread_join](https://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_join.html), which "suspend[s] execution of the [main] thread until the [joinable] thread terminates," which is all the `while` loop is doing.

Ah, OK, you're right.  Though including the extra check would have made correctness more obvious. :-)

I tested this version on the full set of systems listed above, and it behaves as expected:

1. Without the fix, the test fails on 10.4-10.5 and 10.9-10.10, while passing otherwise.
2. With the fix, it passes in all cases.

So we should be good to go except for swapping the commit order.

MarcusCalhoun-Lopez:
Thank you for the testing.
I changed the order of the commits and removed the draft status.
As soon as these changes are in a released version of legacy-support, we can merge PR to update Rust (https://github.com/macports/macports-ports/pull/23874).

fhgwright:
@MarcusCalhoun-Lopez 
> I changed the order of the commits and removed the draft status.

And just to be maximally paranoid, I verified that the end result is identical in both cases. :-)

> As soon as these changes are in a released version of legacy-support, we can merge PR to update Rust ([macports/macports-ports#23874](https://github.com/macports/macports-ports/pull/23874)).

@mascguy 
As usual, after merging this PR we'll first make an updated `legacy-support-devel`, then allow a bit of time for interested parties to test it, then cut a new release and update `legacy-support`.

fhgwright:
@MarcusCalhoun-Lopez @mascguy 
Since this is still awaiting a merge, I additionally tested it with a temporary Portfile (on all the usual systems), with the expected results.

mascguy:
> @MarcusCalhoun-Lopez @mascguy 
> Since this is still awaiting a merge, I additionally tested it with a temporary Portfile (on all the usual systems), with the expected results.

Great, I will merge ASAP.

As for `legacy-support-devel`, feel free to go ahead and update. (Would take care of it, but presently away from my Mac.)

MarcusCalhoun-Lopez:
Would it be possible to crate a new release version of legacy support so that Rust can build on all platforms?

fhgwright:
@MarcusCalhoun-Lopez
> Would it be possible to crate a new release version of legacy support so that Rust can build on all platforms?

@mascguy 
Yeah, the -devel version has had a week of soak time by now, which should be plenty.  I'm working on the security wrapper for `stpncpy()`, but there's no reason to hold up the release for that.  If you cut the release, I can create and test the PR for the port update.

Was "crate" a Freudian slip? :-)

PR #87, opened 2024-05-27
Add implementation and tests for stpncpy() security wrapper

fhgwright:
See the two commit messages.

Tested on:
```
macOS 12.7.5 21H1222, x86_64, Xcode 14.2 14C18
macOS 12.7.5 21H1222, arm64, Xcode 14.2 14C18
macOS 13.6.7 22G720, arm64, Xcode 15.2 15C500b
macOS 14.5 23F79, arm64, Xcode 15.4 15F31d
```

### DIFF
--- a/test/test_stpncpy_chk.c
+++ b/test/test_stpncpy_chk.c
@@ -1,0 +1,209 @@
+/*
+ * Copyright (c) 2024 Frederick H. G. Wright II <fw@fwright.net>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+/*
+ * This provides test cases for the stpncpy() security wrapper.
+ * There may or may not be compile-time warnings in some cases, but
+ * the final pass/fail should be as expected.
+ *
+ * Informational messages regarding the relevant compile flags are also
+ * included.
+ */
+
+/* Allow deferring the stdio.h include */
+int printf(const char *format, ...);
+
+#define PRINT_VAR(x) printf("    %s = %lld\n", #x, (long long) x)
+#define PRINT_UNDEF(x) printf("    " #x " is undefined\n")
+
+void
+print_before(void)
+{
+  printf("  Before string.h include:\n");
+  #ifdef _FORTIFY_SOURCE
+  PRINT_VAR(_FORTIFY_SOURCE);
+  #else
+  PRINT_UNDEF(_FORTIFY_SOURCE);
+  #endif
+  #ifdef _USE_FORTIFY_LEVEL
+  PRINT_VAR(_USE_FORTIFY_LEVEL);
+  #else
+  PRINT_UNDEF(_USE_FORTIFY_LEVEL);
+  #endif
+}
+
+#include <string.h>
+
+void
+print_after(void)
+{
+  printf("  After  string.h include:\n");
+  #ifdef _FORTIFY_SOURCE
+  PRINT_VAR(_FORTIFY_SOURCE);
+  #else
+  PRINT_UNDEF(_FORTIFY_SOURCE);
+  #endif
+  #ifdef _USE_FORTIFY_LEVEL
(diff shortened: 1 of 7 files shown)


fhgwright:
@mascguy 
Adding notification manually.

kencu:
Getting closer now…

but just put it in the expected header.

kencu:
And….why all the previous drama and arguing?

fhgwright:
@kencu 
> Getting closer now…
> 
> but just put it in the expected header.

If by "expected header" you mean an LS `secure/_string.h` (currently nonexistent), well, let's consider that:

First of all, a header of that name would shadow the SDK copy, so it would need to use `#include_next` to get around that.  So far, so good.  But it would be applied as part of the `#include_next <string.h` in the LS `string.h`, and therefore _before_ the declaration of `stpncpy()`.  That would reintroduce the same old build failure, but now not limited to the SDK mismatch case.  So it would need to play some games with the guard macros _and_ get included a second time at the bottom of LS `string.h`, in order to happen at the right time.

The above could be simplified by changing the name of the directory and/or file so that it didn't collide with the SDK `secure/_string.h`.  But then it's no longer exactly mimicking the SDK layout, so it just boils down to the question of whether it makes sense to split `string.h` into two files or keep it as one.  It's logically a single header in either case, since including `string.h` implies including the secondary header (in the relevant cases), and not only is including the secondary header directly  explicitly documented as illegal, but also there's even code to generate an error if you try.  The only thing that matters to the end user is whether `#include <string.h>` does the right thing for the chosen (or defaulted) value of `_FORTIFY_SOURCE`.

The only plausible explanation for why the normal practice is to split the header into two parts is the reduction of clutter in the main header (particularly since the whole security-wrapper thing was an afterthought).  But that's actually a mixed blessing, since it splits the information regarding a given function into two separate pieces in two separate files, thus obfuscating the full picture.  It's further obfuscated by the fact that the `#include <secure/_string.h>` happens at the _bottom_ of `string.h`, which is not where headers including other headers usually happens.  Doing it last is of course a necessity in this case in order to do things in the correct order, but nevertheless, having an additional `#include` in an unusual place adds to the obfuscation.

Also note that this whole issue in the LS context is for a grand total of one function, making the "clutter" fairly minimal.  So the question is whether to take the simple and more understandable approach of putting everything in one header, or taking the more complicated and more obtuse approach of splitting it in two, just for the questionable goal of kinda sorta mimicking the  SDK header layout.

Verdict: One file.

It might surprise you to learn that I thought about all of this before implementing anything.  And just that alone took well over "five minutes". :-) As did explaining it.

BTW, Apple added the source code for `stpcpy()` and `stpncpy()` to their codebase at the same time, at least at the commit resolution of their published codebase.  So they apparently just forgot to "publish" `stpncpy()` until 10.7.

@kencu 
> And….why all the previous drama and arguing?

Because implementing, debugging, and testing the enhancement, plus implementing, debugging, and testing the test for the enhancement was _massively_ more effort than the one-line addition sufficient to fix the actual bug.  _That_ fix is already deployed and working.

The whole "five minutes" thing was clearly based on a variant of Weiler's law.

@mascguy 
Aren't you glad now that _you_ didn't implement `stpncpy()`? :-)

test/test_stpncpy_chk.c
```diff
@@ -0,0 +1,209 @@
...
+  PRINT_UNDEF(_FORTIFY_SOURCE);
+  #endif
+  #ifdef _USE_FORTIFY_LEVEL
+  PRINT_VAR(_USE_FORTIFY_LEVEL);
+  #else
+  PRINT_UNDEF(_USE_FORTIFY_LEVEL);
+  #endif
+}
+
+#if defined(_USE_FORTIFY_LEVEL) && _USE_FORTIFY_LEVEL > 1
```
mascguy, line 66: In `string.h`, checks appear to be enabled when `_USE_FORTIFY_LEVEL` > 0.

But here, we're only enabling tests when the value is greater than 1.

Admittedly my knowledge of these preproc symbols is lacking. (Particularly in terms of the semantics related to their values.)

But shouldn't tests be enabled, if the checks are...?

mascguy (approved):
Apart from my one question, this looks great! Everything is well-documented, logically organized, etc.

And while I haven't tested with it yet, there are no concerns here.

Anyone else? @kencu and @MarcusCalhoun-Lopez ?

test/test_stpncpy_chk.c
fhgwright: You're right, though it's not clear when it makes a difference.  I had '2' on the brain because that's the only value that ever gets set by default.  I'm not sure what the difference is, if any, between '1' and '2'.  But for consistency, I changed the threshold in the test and added another "forced" test setting it to '1' for good measure.  And retested on all the above systems.

I didn't retest again after fixing the typo in the Makefile comment. :-)

mascguy: > But for consistency, I changed the threshold in the test and added another "forced" test setting it to '1' for good measure.  And retested on all the above systems.

Sounds good!

fhgwright:
@mascguy 
Ping.

I have other changes coming soon, so it may not make sense to be in a hurry to update the -devel port.

mascguy:
> I have other changes coming soon, so it may not make sense to be in a hurry to update the -devel port.

I'm fine with either merging ASAP, or deferring until your other changes are done. Up to you!

fhgwright:
@mascguy 
> > I have other changes coming soon, so it may not make sense to be in a hurry to update the -devel port.
> 
> I'm fine with either merging ASAP, or deferring until your other changes are done. Up to you!

Those two things are separate.  I'd like to get _this_ PR merged, since there are lexical conflicts with other things I'm working on, but there's no rush for a companion update to the -devel port.

Unless someone is really champing at the bit to get security wrappers for C-only uses of `stpncpy()` on 10.6. :-)

mascguy:
> Those two things are separate. I'd like to get _this_ PR merged, since there are lexical conflicts with other things I'm working on, but there's no rush for a companion update to the -devel port.

Got it, I'll go ahead and merge then. Thanks for clarifying!

PR #93, opened 2024-07-15
Add fgetattrlist()

macportsraf:
This adds an implementation of `fgetattrlist()`. Tested on 10.14.6, 10.6.8, and 10.4.11 (ppc) but only the 10.4 test means anything. The test is very basic. It just calls the function on the current directory's file descriptor, and only requires that it doesn't return an error.

I noticed a strange thing when testing on 10.4 (ppc). The first time I run:

    PATH=/opt/local/bin:$PATH CC=gcc-apple-4.2 CFLAGS=-isystem`pwd`/tiger_only/include make test

I get an error:

    ld: file not found: src/fdopendir.dl.o.inode32.ppc

But if I run it again, it works. This happens with or without this change. I suspect that it doesn't matter.

### DIFF
--- a/include/MacportsLegacySupport.h
+++ b/include/MacportsLegacySupport.h
@@ -222,7 +222,7 @@
 /*  realpath() on < 10.6 does not support modern NULL buffer usage */
 #define __MPLS_LIB_SUPPORT_REALPATH_WRAP__    (__MPLS_TARGET_OSVER < 1060)
 
-/* setattrlist */
+/* fsetattrlistat, fgetattrlistat */
 #define __MPLS_SDK_SUPPORT_FSETATTRLIST__     (__MPLS_SDK_MAJOR < 1060)
 #define __MPLS_LIB_SUPPORT_FSETATTRLIST__     (__MPLS_TARGET_OSVER < 1060)
 
--- a/src/fgetattrlist.c
+++ b/src/fgetattrlist.c
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2024 Mihai Moldovan <ionic@ionic.de>, raf <raf@raf.org>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+/* MP support header */
+#include "MacportsLegacySupport.h"
+
+#if __MPLS_LIB_SUPPORT_FSETATTRLIST__
+
+#include <sys/attr.h>
+#include <sys/fcntl.h>
+#include <sys/param.h>
+#include <sys/errno.h>
+
+#include <fcntl.h>
+#include <string.h>
+#include <unistd.h>
+
+
+#ifdef __LP64__
+int fgetattrlist(int fd, void *a, void *buf, size_t size, unsigned int flags)
+#else /* defined (__LP64__) */
+int fgetattrlist(int fd, void *a, void *buf, size_t size, unsigned long flags)
+#endif /* defined (__LP64__) */
+{
+    int cont = 1,
+        ret = 0;
+
+    char fpath[MAXPATHLEN];
+    memset (fpath, 0, MAXPATHLEN);
+    if (-1 == fcntl(fd, F_GETPATH, fpath)) {
+        ret = EBADF;
+        cont = 0;
+    }
+
+#if __MPLS_TARGET_OSVER < 1080
+    /*
+     * Older systems don't correctly check if no attributes are to be read, which usually
+     * means a buffer size of zero and return an error since they malloc a block of
+     * memory with size zero, leading to ENOMEM.
+     *
+     * Emulate the fix from 10.8 for those.
+	 *
+	 * Note: This check is inherited from fsetattrlist() and might or might
+	 * not be appropriate here (but it seems reasonable).
+     */
+    const struct attrlist *al = a;
+    if (al->commonattr == 0 &&
+        (al->volattr & ~ATTR_VOL_INFO) == 0 &&
+        al->dirattr == 0 &&
+        al->fileattr == 0 &&
+        al->forkattr == 0) {
+        cont = 0;
+
+        /*
+         * Explicitly let the potential error from above pass through, since that's what
+         * the original function seems to do as well.
+         */
+    }
+#endif
+
+    if (cont) {
+        ret = getattrlist(fpath, a, buf, size, flags);
+    }
+
+    return ret;
+}
+
+#endif  /* __MPLS_LIB_SUPPORT_FSETATTRLIST__ */
--- a/README.md
+++ b/README.md
@@ -199,7 +199,8 @@ Wrapped headers and replaced functions are:
     <td rowspan="3"><code>sys/unistd.h</code></td>
     <td>Adds <code>getattrlistat</code>, <code>readlinkat</code>, <code>faccessat</code>,
         <code>fchownat</code>, <code>linkat</code>, <code>symlinkat</code>,
-        and <code>unlinkat</code> functions</td>
+        <code>unlinkat</code></td>,
+        <code>fsetattrlist</code>, and <code>fgetattrlist</code> functions.
     <td>OSX10.9</td>
   </tr>
   <tr>
--- a/include/unistd.h
+++ b/include/unistd.h
@@ -101,8 +101,10 @@
 
 #ifdef __LP64__
 int   fsetattrlist(int,void*,void*,size_t,unsigned int);
+int   fgetattrlist(int,void*,void*,size_t,unsigned int);
 #else /* defined (__LP64__) */
 int   fsetattrlist(int,void*,void*,size_t,unsigned long);
+int   fgetattrlist(int,void*,void*,size_t,unsigned long);
 #endif /* defined (__LP64__) */
 
 #endif  /* __MPLS_SDK_SUPPORT_FSETATTRLIST__ */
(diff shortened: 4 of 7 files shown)


fhgwright:
A few notes:

1. I don't think it's necessary to have separate feature flags for `fgetattrlist()` and `fsetattrlist()`, given that they appeared at the same time and are documented as going hand-in hand.  It may or may not make sense in the spirit of egalitarianism to rename the feature to `FXETATTRLIST`. :-) A similar issue might apply to the sources.

2. PR #92 makes substantial changes that interact with this, so you should look at that, or wait until it's merged to finalize this one.  Though sharing the feature flag(s) from `fsetattrlist()` _without_ a rename might not conflict.

3. This might be a good opportunity to add a test for `fsetattrlist()`, which was previously neglected.

4. There's a subtle potential problem in that these functions aren't supported by all filesystems.  It shouldn't be an issue for `HFS+` or `APFS`, but if someone tried to run the test on a weird filesystem it might fail.  I don't think it's worth treating that as a pass, but it probably makes sense to provide a more informative error in that case.  I think all that's needed is to try a `getattrlist()` and check the error code.  I expect that the failure can be seen by using a FAT32 partition on a USB stick.

fhgwright:
> I noticed a strange thing when testing on 10.4 (ppc). The first time I run:
> 
> ```
> PATH=/opt/local/bin:$PATH CC=gcc-apple-4.2 CFLAGS=-isystem`pwd`/tiger_only/include make test
> ```
> 
> I get an error:
> 
> ```
> ld: file not found: src/fdopendir.dl.o.inode32.ppc
> ```
> 
> But if I run it again, it works. This happens with or without this change. I suspect that it doesn't matter.

That sounds like a missing dependency in the Makefile, though I haven't seen it myself.

BTW, if you're working from the latest `master` and not the release version, then `tiger-only/include` no longer exists, and no special setup is needed for 10.4 (other than the compiler choice).

fhgwright:
A couple of other notes:

1. I think it's best to put implementations and tests into separate commits, to facilitate reproducing the test failure when the implementation is missing.  The implementation should come first, to avoid creating a "pothole" in the history.
2. This should probably be a draft until #92 is merged, and this one is adjusted as needed.

macportsraf:
> A few notes:
> 
>     1. I don't think it's necessary to have separate feature flags for `fgetattrlist()` and `fsetattrlist()`, given that they appeared at the same time and are documented as going hand-in hand.  It may or may not make sense in the spirit of egalitarianism to rename the feature to `FXETATTRLIST`. :-) A similar issue might apply to the sources.

I've removed `__MP_LEGACY_SUPPORT_FGETATTRLIST__`.

>     2. PR [Mostly SDK-related changes, including __DARWIN_C_LEVEL stuff #92](https://github.com/macports/macports-legacy-support/pull/92) makes substantial changes that interact with this, so you should look at that, or wait until it's merged to finalize this one.  Though sharing the feature flag(s) from `fsetattrlist()` _without_ a rename might not conflict.

I'm happy to let others decide that.

>     3. This might be a good opportunity to add a test for `fsetattrlist()`, which was previously neglected.

Added. The test needed `fgetattrlist()` (taken from the first code sample in the manual entry), so maybe the other `fgetattrlist()` (taken from its manual entry) is no longer necessary, but it's harmless. 

>     4. There's a subtle potential problem in that these functions aren't supported by all filesystems.  It shouldn't be an issue for `HFS+` or `APFS`, but if someone tried to run the test on a weird filesystem it might fail.  I don't think it's worth treating that as a pass, but it probably makes sense to provide a more informative error in that case.  I think all that's needed is to try a `getattrlist()` and check the error code.  I expect that the failure can be seen by using a FAT32 partition on a USB stick.

I've added some output in the case of test failure to state that it might be correctly failing due to an unexpected filesystem type.

macportsraf:
> That sounds like a missing dependency in the Makefile, though I haven't seen it myself.

It was just a time error on the laptop. For some reason, it doesn't set the time until I go to the system preferences and look at the time settings.

> BTW, if you're working from the latest `master` and not the release version, then `tiger-only/include` no longer exists, and no special setup is needed for 10.4 (other than the compiler choice).

Thanks.

macportsraf:
> A couple of other notes:
> 
>     1. I think it's best to put implementations and tests into separate commits, to facilitate reproducing the test failure when the implementation is missing.  The implementation should come first, to avoid creating a "pothole" in the history.

Done.

>     2. This should probably be a draft until [Mostly SDK-related changes, including __DARWIN_C_LEVEL stuff #92](https://github.com/macports/macports-legacy-support/pull/92) is merged, and this one is adjusted as needed.

It'll be here when you want it.

include/MacportsLegacySupport.h
```diff
@@ -186,7 +186,7 @@
 /*  realpath() on < 1060 does not support modern NULL buffer usage */
 #define __MP_LEGACY_SUPPORT_REALPATH_WRAP__   (__MPLS_TARGET_OSVER < 1060)
 
-/* setattrlistat */
+/* fsetattrlistat */
```
fhgwright, line 189: Should mention both functions now.

src/fgetattrlist.c
```diff
@@ -0,0 +1,80 @@
...
+    char fpath[MAXPATHLEN];
+    memset (fpath, 0, MAXPATHLEN);
+    if (-1 == fcntl(fd, F_GETPATH, fpath)) {
+        ret = EBADF;
+        cont = 0;
+    }
+
+#if __MPLS_TARGET_OSVER < 1080
+    /*
+     * Older systems don't correctly check if no attributes are to be set, which usually
```
fhgwright, line 49: Copy-'n-paste error: "set" isn't really correct here.  Maybe "read" or "obtained".

Based on the description, I don't know if the problem addressed here exists on reads at all, though as you say, the "fix" probably doesn't hurt.

fhgwright:
Note that since this interacts with `__DARWIN_C_LEVEL`, it will also need the obvious one-line addition to `manual_tests/darwin_c.c`, which won't appear until #92 is merged.  This is mentioned in the comment in `MacportsLegacySupport.h` as (to be) updated.  "Need" is in the sense of "for completeness of the test", not "for correct functioning".

That's the trouble with adding features while the framework is in flux. :-)

`README.md` will need an update as well.

include/MacportsLegacySupport.h
macportsraf: Done

src/fgetattrlist.c
macportsraf: Done

macportsraf:
> Note that since this interacts with `__DARWIN_C_LEVEL`, it will also need the obvious one-line addition to `manual_tests/darwin_c.c`, which won't appear until #92 is merged. This is mentioned in the comment in `MacportsLegacySupport.h` as (to be) updated. "Need" is in the sense of "for completeness of the test", not "for correct functioning".
> 
> That's the trouble with adding features while the framework is in flux. :-)

Of course, this patch could go in first, and the other patch could be rebased to take it into account. But I assume there must be good reasons not to do that. I don't understand why this patch would have anything to do with manual tests, since the tests aren't manual, but that's OK. I probably don't need to understand. Just let me know what change to make to `manual_tests/darwin_c.c` when the time comes.

> `README.md` will need an update as well.

Done. `fsetattrlist` wasn't in there either but it is now.

fhgwright:
@macportsraf 
> Of course, this patch could go in first, and the other patch could be rebased to take it into account. But I assume there must be good reasons not to do that. I don't understand why this patch would have anything to do with manual tests, since the tests aren't manual, but that's OK. I probably don't need to understand. Just let me know what change to make to `manual_tests/darwin_c.c` when the time comes.

Well, #92 came first and is much more extensive.  And once you see it, you'll see what the obvious addition is for `fgetattrlist()`.

The `darwin_c` tests are only manual for a weird reason involving the Makefile.

> > `README.md` will need an update as well.
> 
> Done. `fsetattrlist` wasn't in there either but it is now.

Ah, you're right. I'd missed that.  I'd also missed the header typo until after #92. :-)

include/MacportsLegacySupport.h
```diff
@@ -186,7 +186,7 @@
 /*  realpath() on < 1060 does not support modern NULL buffer usage */
 #define __MP_LEGACY_SUPPORT_REALPATH_WRAP__   (__MPLS_TARGET_OSVER < 1060)
 
-/* setattrlistat */
+/* fsetattrlistat, fgetattrlistat */
```
fhgwright, line 189: There's a trivial lexical conflict here with the updated `master`, due to non-conflicting changes on adjacent lines.

src/fgetattrlist.c
```diff
@@ -0,0 +1,80 @@
...
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+/* MP support header */
+#include "MacportsLegacySupport.h"
+#if __MP_LEGACY_SUPPORT_FSETATTRLIST__
```
fhgwright, line 19: The flag name needs to change to `__MPLS_LIB_SUPPORT_FSETATTRLIST__`, to match the new naming scheme.

Also, I think it's visually better to put a blank line between the include and the conditional.  I already did that  in `fsetattrlist.c`, but it wouldn't have been in `master` at the time you created this PR.

src/fgetattrlist.c
```diff
@@ -0,0 +1,80 @@
...
+#endif
+
+    if (cont) {
+        ret = getattrlist(fpath, a, buf, size, flags);
+    }
+
+    return ret;
+}
+
+#endif  /* __MP_LEGACY_SUPPORT_FSETATTRLIST__ */
```
fhgwright, line 80: Same name change here.

fhgwright:
@macportsraf 
With the above changes to the feature commit, and the previously mentioned one-line addition to the test commit, this will be properly referenced to the new `master`.  I already tested a local version with those changes.

macportsraf:
Thanks. It's updated.

fhgwright:
@macportsraf 
> Thanks. It's updated.

Almost. :-)

As previously mentioned, the test commit should include this:
```
--- a/manual_tests/darwin_c.c
+++ b/manual_tests/darwin_c.c
@@ -212,6 +212,7 @@ int timespec_get = 0;
 #ifdef _SC_PHYS_PAGES
 #error _SC_PHYS_PAGES is unexpectedly defined
 #endif
+int fgetattrlist = 0;
 int fsetattrlist = 0;
 #endif /* __DARWIN_C_LEVEL < __DARWIN_C_FULL */
```

macportsraf:
Thanks. It's there now.

fhgwright:
@macportsraf 
> Thanks. It's there now.

Umm, you didn't actually change it:
```
MacPro:macports-legacy-support fw$ git diff 5d60b93 b34b0bb
MacPro:macports-legacy-support fw$ 
```
Did you forget a `-a` perhaps?

macportsraf:
Oops. It's there now.

fhgwright:
@macportsraf 
> Oops. It's there now.

Yup.

That now matches what I'd already tested, on:
```
macOS 12.7.5 21H1222, x86_64, Xcode 14.2 14C18
macOS 12.7.5 21H1222, arm64, Xcode 14.2 14C18
macOS 13.6.7 22G720, arm64, Xcode 15.2 15C500b
macOS 14.5 23F79, arm64, Xcode 15.4 15F31d
```
That actually understates it, since I also built and ran the tests with all compatible SDKs not earlier than the target OS.

@mascguy 
LGTM

mascguy:
> @mascguy 
> LGTM

Great, I'll go ahead and merge

PR #94, opened 2024-07-21
Add two new queue macros with tests, plus a warning fix.

fhgwright:
See the individual commit messages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,15 @@ Wrapped headers and replaced functions are:
     <td>OSX10.10</td>
   </tr>
   <tr>
-    <td><code>sys/queue.h</code></td>
+    <td rowspan="3"><code>sys/queue.h</code></td>
+    <td>Adds <code>SLIST_HEAD_INITIALIZER</code> macro</td>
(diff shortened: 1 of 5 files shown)


fhgwright:
@mascguy 
This should be a much easier review than the last one. :-)

mascguy:
> @mascguy 
> This should be a much easier review than the last one. :-)

Yes indeed, and looks good to me.

mascguy (approved):

fhgwright:
I'll wait until an appropriately updated #93 is also merged before making another -devel update.  This completes what I'd planned for this release cycle, so the next -devel will probably be an RC.  With the extent of the changes, it might be a good idea to call for wider testing at that point.

mascguy:
> I'll wait until an appropriately updated #93 is also merged before making another -devel update.  This completes what I'd planned for this release cycle, so the next -devel will probably be an RC.  With the extent of the changes, it might be a good idea to call for wider testing at that point.

Sounds good. And agreed, wider testing would be a good thing.

PR #95, opened 2024-08-20
Add vdprintf(), Fix dprintf()

macportsraf:
The `dprintf()` function has a big bug. It closes the file descriptor that is passed to it. So it only works once. After that, there is no output on that file descriptor. This is because it calls `fdopen()` and `fclose()` but it doesn't first call `dup()`. This pull request fixes that, and adds the corresponding `vdprintf()` function as well (tested on 10.6 and 10.4).

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ Wrapped headers and replaced functions are:
   </tr>
   <tr>
     <td><code>stdio.h</code></td>
-    <td>Adds <code>dprintf</code>, <code>getline</code>, <code>getdelim</code>,
+    <td>Adds <code>dprintf</code>, <code>vdprintf</code>, <code>getline</code>, <code>getdelim</code>,
         <code>open_memstream</code>, and <code>fmemopen</code> functions</td>
     <td>OSX10.6, OSX10.12 (open_memstream)</td>
   </tr>
--- a/include/MacportsLegacySupport.h
+++ b/include/MacportsLegacySupport.h
@@ -173,7 +173,7 @@
 #define __MPLS_SDK_SUPPORT_STRNDUP__          (__MPLS_SDK_MAJOR < 1070)
 #define __MPLS_LIB_SUPPORT_STRNDUP__          (__MPLS_TARGET_OSVER < 1070)
 
-/* dprintf */
+/* dprintf, vdprintf */
 #define __MPLS_SDK_SUPPORT_DPRINTF__          (__MPLS_SDK_MAJOR < 1070)
 #define __MPLS_LIB_SUPPORT_DPRINTF__          (__MPLS_TARGET_OSVER < 1070)
 
--- a/include/stdio.h
+++ b/include/stdio.h
@@ -37,6 +37,7 @@
 
 __MP__BEGIN_DECLS
 extern int dprintf(int fd, const char * __restrict format, ...);
+extern int vdprintf(int fd, const char * __restrict format, va_list ap);
 __MP__END_DECLS
 
 #endif /* __MPLS_SDK_SUPPORT_DPRINTF__ */
--- a/manual_tests/darwin_c.c
+++ b/manual_tests/darwin_c.c
@@ -57,6 +57,7 @@ int fdopendir = 0;
 #include <stdio.h>
 #if __DARWIN_C_LEVEL < 200809L
 int dprintf = 0;
+int vdprintf = 0;
 int getdelim = 0;
 int getline = 0;
 int open_memstream = 0;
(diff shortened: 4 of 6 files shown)


fhgwright:
@macportsraf 
I only just noticed this, since there are no automatic notifications for PRs.  I'll test it soon.

fhgwright:
@macportsraf 
First an organizational comment:

When both `dprintf()` and `vdprintf()` are being provided, it makes more sense to make `dprintf()` a wrapper around `vdprintf()`, rather than duplicating the "guts".  That's what Apple's code does.

On the functional part:

Duplicating the `fd` to defang `fclose()` seemed like a kludgy approach, and one that might not match the normal behavior in corner cases, so I looked at what Apple's code does.  It boils down to:

1. Implicitly allocate a FILE as an auto variable, along with a buffer of size BUFSIZ (from stdio.h).
2. Initialize various fields of the FILE, setting it up to use the specified `fd` and the local buffer.
3. Perform the stream output.
4. Use `fflush()` to flush out any buffered data.
5. Implicitly free the FILE and buffer by returning from the function.

So steps 1 and 2 are essentially equivalent to `fdopen()` while using purely local storage, and there's no `fclose()` (though the `fflush()` that's the important part of `fclose()` is present.  The problem with trying to use this approach in `legacy-support` is that step 2 involves some internal knowledge of a FILE that's not safely available to "outsiders".  But I think it can be made to work as follows:

1. Implicitly allocate a buffer as an auto variable (using BUFSIZ from stdio.h).
2. Use `fdopen()` to allocate and initialize a new FILE (as in the current code, without the `dup()`).
3. Use `setbuffer()` to set up FILE to use the local buffer rather than one from `malloc()`.
4. Perform the stream output.
5. Use `fflush()` to flush out any buffered data.
6. Explicitly free the FILE with `free()`.
7. Implicitly free the buffer by returning from the function.

This should come pretty close to what Apple's code does, except for needing to `malloc()/free()` the FILE.

fhgwright:
@macportsraf 
A few additional notes:

1. Since `vdprintf()` (like `dprintf()`) is affected by `__DARWIN_C_LEVEL`, it should be added to `darwin_c.c`.  That test will be moving in an upcoming change, but if this PR is updated and merged first you can ignore that.
2. The test should probably include a normal stdout use after the current two, to prove that the close bug is truly fixed.  Granted, the use of `vdprintf()` after `dprintf()` proves that `dprintf()` is OK, but it doesn't prove that `vdprintf()` is OK.
3. As usual, I recommend putting implementations and tests in separate commits, to facilitate potential reconstruction of the failing tests in the future by simple commit-picking.  In this case, it's a bit more complicated, since it's both fixing an existing function and adding a new one, and a test of both without the "fix" would fail to build and thus not demonstrate the runtime problem.  I don't think it's necessary to separate the `vdprintf()` addition from the `dprintf()` fix, especially since they might interact (as recommended above), but splitting the test would make sense.
 
Hence, I'd recommend:

1. One commit to add `vdprintf()` and fix `dprintf()`.
2. One commit to add a test for `dprintf()` only.
3. One commit to add `vdprintf()` to the `dprintf()` test.

It's not critical exactly where the `darwin_c.c` addition fits here, since it's independent of the implementation.  It's just a test of the headers, and it's perfectly valid for functions not provided by `legacy-support` at all.

fhgwright:
@macportsraf 
If you don't have the time or the inclination to mess with this further, I could take it over.  I'm working on some other `legacy-support` updates anyway.

macportsraf:
I'm still waiting for my main laptop with the slowly exploding batteries
(i.e. scarily swollen) batteries to come back from the Apple repair people.
So feel free to take it over. Thanks.

fhgwright:
@macportsraf 
> I'm still waiting for my main laptop with the slowly exploding batteries (i.e. scarily swollen) batteries to come back from the Apple repair people. So feel free to take it over. Thanks.

OK, will do.  This PR should probably be closed, then.

It looks like the nasty close bug has been there for almost three years, so it's not exactly an emergency fix. :-)

fhgwright:
> @macportsraf A few additional notes:
> 
> 1. Since `vdprintf()` (like `dprintf()`) is affected by `__DARWIN_C_LEVEL`, it should be added to `darwin_c.c`.  That test will be moving in an upcoming change, but if this PR is updated and merged first you can ignore that.

Sorry, I'd missed that you'd already done that.  It's moot now, of course.

fhgwright:
@macportsraf 
So it looks like I was half-right on the use of `dup()`.

I was wrong about being able to apply `free()` to the `FILE`, making `fclose()` the only known way to free it, and needing `dup()` to avoid closing the original `fd`.

But I was right about the use of `dup()` not being entirely correct, because `dup()` doesn't duplicate locks, so the implied `close()` frees any locks that were originally held.  This is still a bug, albeit a less serious one than closing the `fd` altogether, so that's the way I currently have the code.  I'm still using the buffer on the stack for efficiency.

PR for this and other changes coming soon, after more extensive testing.

PR #97, opened 2024-10-29
Many updates - see individual commits

fhgwright:
Notably:

- Mismatched SDK compatibility is extended to include 15.x SDK.
- Broken 10.4 CLOCK_MONOTONIC is fixed.
- Empty modules are now (mostly) excluded from static library.
- Tests are now provided for static and system-replacement libraries.
- Parallel builds are fixed.

Tested on:
```
macOS 12.7.6 21H1320, x86_64, Xcode 14.2 14C18
macOS 12.7.6 21H1320, arm64, Xcode 14.2 14C18
macOS 13.7 22H123, arm64, Xcode 15.2 15C500b
macOS 14.7 23H124, arm64, Xcode 16.0 16A242d
macOS 15.0.1 24A348, arm64, Xcode 16.0 16A242d
```


fhgwright:
@mascguy 
Here's another round of changes.  There are a few more that I thought I could include, but I ran into a weird problem in final testing, so those will have to wait.  Everything here is fully tested.

The `-devel` port update for this will be able to omit the parallel-build disable, though not for the main port until the next release.  I should probably do that one so that it can be fully tested.

@MarcusCalhoun-Lopez 
Perhaps you could comment on the alleged but undocumented need for the `os_unfair_lock` test to be in C++.

@macportsraf
Perhaps you could comment on the assumed obsoleteness of `dirent_with_cplusplus`.

For both of the latter, see `test/README-CPP.txt`, added by `e17a364`.

macportsraf:
Agreed. Since the `seekdir` macro is no longer there, there's no need for that test.

mascguy:
While I don't have any concerns with the changes, it would probably be good to have at least one other person review them as well: Given how much is included, it's challenging to scrutinize it all. (Definitely not a criticism though, as it's all good stuff.)

mascguy (approved):

fhgwright:
@mascguy 
> While I don't have any concerns with the changes, it would probably be good to have at least one other person review them as well: Given how much is included, it's challenging to scrutinize it all. (Definitely not a criticism though, as it's all good stuff.)

I don't have any problem with more reviewers, as long as it can be done in a timely fashion.  And I specifically requested feedback from @MarcusCalhoun-Lopez regarding the inconsistency between the comments and the reality for the `os_unfair_lock` test (not to mention the ticket I filed to get an explanation of the `atexit` mess).

Getting _too_ carried away with reviewers can have its problems, though.  One time at Google I needed to make a change that both had the potential to screw up thousands of machines and was sufficiently time-critical that it couldn't afford to wait for the normal graduated rollout process.  The manager made having several experienced reviewers a condition for short-circuiting the rollout, but then the code review degenerated into arguments among the reviewers themselves over style rather than substance. :-)

mascguy:
Fred, do you need feedback from @MarcusCalhoun-Lopez before we merge?

It would be great to get these changes out to a small group via `legacy-support-devel`, but there's no rush. Let me know your thoughts.

fhgwright:
@mascguy 
> Fred, do you need feedback from @MarcusCalhoun-Lopez before we merge?

It's not essential, but I thought it would be useful to get clarification on the comment that the `os_unfair_lock` test needs to be in C++ while referencing something that makes no such claim.  I didn't actually remove the C++ test, but I relegated it to the manual category (for stated reasons) while replacing it with a C version.  So at worst, it just means that the allegedly necessary C++ test is no longer part of the normal `test` target, but is still available.  It still passes in the cases where I've run it, but some target/SDK combinations involving SDK 15 fail to build it.

> It would be great to get these changes out to a small group via `legacy-support-devel`, but there's no rush. Let me know your thoughts.

If you just mean creating a `legacy-support-devel` with this code version, then of course that was always intended.  But as you know, for long-term consistency, the changes need to be merged in the project _before_ updating `legacy-support-devel`.

If you mean issuing a call for more widespread testing via `legacy-support-devel`, that could of course be done once it's been updated, although in general I probably wouldn't bother doing that on the ML until it's reached the RC state, and this isn't intended to be the last update in this cycle.

I also belatedly realized that the renaming I did to the overly verbose filenames would require updating the `atexit` deletion kludge in the Portfile.  Rather than requiring the kludge to be modified and then subsequently removed, I was planning (as soon as I've tested it today) to add one more commit to this PR to move the `atexit` source to a different directory to hide it without resorting to the delete kludge in the Portfile (and hence skipping the intermediate adjustment in the Portfile).

The history of the `atexit` wrapper (as yet to be elaborated upon by Marcus) seems to be that:

1. It was originally introduced in Apr-2022.
2. It was almost immediately removed via the Portfile kludge, since it caused trouble.
3. It was never even disabled in a more reasonable way, let alone fixed to work properly.

Moving the file in the repo instead of deleting it in the Portfile (which just maintains the status quo) means that the Portfile-based build and the bare build in the repo will have the same end result for the first time on over 2.5 years (for the platforms where the wrapper is applicable).

fhgwright:
@mascguy 
I wound up adding two commits.  The first is the `atexit.c` move that I'd mentioned.  The second is a fix for an intermittent `Makefile` issue that I'd seen but thought to be harmless, but which turned out to cause intermittent test failures in the MacPorts build environment.  It didn't seem serious enough to retroactively amend the original commit that introduced it, so I just added a new commit.

The updated PR is fully tested on all platforms, including with a locally modified Portfile.

mascguy (approved):

fhgwright:
@mascguy 
Thanks.  I'll make and test a -devel update today.

PR #98, opened 2024-11-07
Multiple updates, mostly for realpath()

fhgwright:
See the individual commit messages.

Highlights:

- Rewrites realpath() wrapper to handle its variants more correctly.
- Expands realpath() tests to cover all variants.


fhgwright:
@mascguy 
Here's another set of changes.  I'm planning to add a fix for the 10.6 non-POSIX bug, but I ran out of time to get this out before becoming mostly unavailable for about a week.  As is, it's fully tested, including via a temporary Portfile.

mascguy (approved):

PR #99, opened 2024-11-08
Additional realpath() fixes

fhgwright:
- Fix 10.6 non-POSIX realpath()
- Add test case demonstrating 10.6 non-POSIX bug.


fhgwright:
@mascguy 
I had time to implement the other `realpath` fix and its test, and test it.  I don't expect to do anything further before next midweek.  It's probably not super-urgent to get this into a -devel port, but I did test it in the Portfile context just in case you want to do it.  This time it's just the usual commit/version/checksum update.

The only remaining known `realpath` bug is in the non-Darwin-extension (and non-default) 10.15 version (an OS bug, not a `legacy-support` bug), and I wasn't planning to deal with that one right away.

mascguy (approved):

PR #104, opened 2024-11-28
Delete disabled directory

copyleft21:


fhgwright:
@copyleft21 
There is a reason that this directory and its contents are there, which you'd know if you'd bothered to read the history.

mascguy:
> @copyleft21 There is a reason that this directory and its contents are there, which you'd know if you'd bothered to read the history.

Based on the frivolous commits to various repos, this is a spam account. Ignore it.

fhgwright:
@mascguy 
> Based on the frivolous commits to various repos, this is a spam account. Ignore it.

Yeah, and I wasn't too impressed with their other two PRs, either. :-) But this one shows enough attention to the code to know that removing that directory wouldn't break anything.  Maybe they're testing the waters.

I'm not a big fan of anonymous contributors in general.  There are certainly contexts where being anonymous is quite important, but I don't see contributing to MacPorts as one of them.  One would at least hope that anonymous contributions would get an extra degree of scrutiny, lest they become the next `xz`.

PR #112, opened 2025-03-23
uninitialize mode in openat calls

Un1q32:
AFAIK if `O_CREAT` isn't in flags, the mode argument is ignored.  Leaving the variable uninitialized saves the compiler from having to generate an instruction to initialize it to 0 when `O_CREAT` isn't specified.  This should save exactly 1 instruction on most architectures.  The micro-est of micro-optimizations, but still an objective improvement.

### DIFF
--- a/src/atcalls.c
+++ b/src/atcalls.c
@@ -204,7 +204,7 @@ int mknodat(int dirfd, const char *pathname, mode_t mode, dev_t dev)
 
 int openat(int dirfd, const char *pathname, int flags, ...)
 {
-    mode_t mode = 0;
+    mode_t mode;
     if (flags & O_CREAT) {
         va_list ap;
         va_start(ap, flags);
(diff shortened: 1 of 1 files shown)


mascguy:
@fhgwright What do you think?

fhgwright:
Seriously?

First of all, saving one instruction on a call that takes hundreds, if not thousands of instructions isn't even a "micro-est of micro-optimization", but more like a femto-optimization.

Secondly, some compilers aren't too bright about determining what variables "may be used uninitialized", and the easiest fix is usually just to throw in a theoretically unnecessary initializer.  This code is expected to compile without warnings on a very wide range of compilers, which I actually test.

Given the nature of the change and the nature of the userid, I wonder if this is just another spammy proposal, like the previously proposed deletion of the "disabled" directory.

mascguy:
Sounds good, closing

PR #124, opened 2025-07-26
fix overflow in get_thread_usage_(ns/ts) functions

Un1q32:
You are using the SYS_thread_selfusage syscall, which returns a uin64_t, but the syscall function has a return type of int, this causes the return value to overflow at 2^31, and since this will only run on 10.10 or 10.11, where x86 is the only supported platform, and mach time scales directly to nanoseconds, that means this breaks for threads that run longer than slightly more than 2 seconds.  We can get around that by declaring our own function pointer with a uint64_t return type and using dlfcn to get a function pointer, allowing us to override the return type and access the full 64 bits.

### DIFF
--- a/src/time.c
+++ b/src/time.c
@@ -695,7 +695,7 @@ get_thread_usage_ns(void)
 {
   static uint64_t (*syscall64)(int, ...) = NULL;
   if (!syscall64)
-    syscall64 = dlfcn(RTLD_NEXT, "syscall");
+    syscall64 = dlsym(RTLD_NEXT, "syscall");
 
   uint64_t mach_time = syscall64(SYS_thread_selfusage);
 
@@ -708,7 +708,7 @@ get_thread_usage_ts(struct timespec *ts)
 {
   static uint64_t (*syscall64)(int, ...) = NULL;
   if (!syscall64)
-    syscall64 = dlfcn(RTLD_NEXT, "syscall");
+    syscall64 = dlsym(RTLD_NEXT, "syscall");
 
   uint64_t mach_time = syscall64(SYS_thread_selfusage);
 
--- a/src/time.c
+++ b/src/time.c
@@ -690,13 +690,12 @@ get_thread_usage_ts(struct timespec *ts)
  *
  * We do a trick here to override the return type of syscall to avoid overflows.
  */
+
+extern uint64_t syscall64(int, ...) __asm("_syscall");
+
 static inline uint64_t
 get_thread_usage_ns(void)
 {
-  static uint64_t (*syscall64)(int, ...) = NULL;
-  if (!syscall64)
-    syscall64 = dlsym(RTLD_NEXT, "syscall");
-
   uint64_t mach_time = syscall64(SYS_thread_selfusage);
 
   return mach2nanos(mach_time);
@@ -706,10 +705,6 @@ get_thread_usage_ns(void)
 static inline int
 get_thread_usage_ts(struct timespec *ts)
 {
-  static uint64_t (*syscall64)(int, ...) = NULL;
-  if (!syscall64)
-    syscall64 = dlsym(RTLD_NEXT, "syscall");
-
   uint64_t mach_time = syscall64(SYS_thread_selfusage);
 
   mach2timespec(mach_time, ts);
--- a/src/time.c
+++ b/src/time.c
@@ -685,11 +685,19 @@ get_thread_usage_ts(struct timespec *ts)
 
 #else /* __MPLS_TARGET_OSVER >= 101000 */
 
-/* Get the CPU usage of the current thread via syscall, in nanoseconds. */
+/*
+ * Get the CPU usage of the current thread via syscall, in nanoseconds.
+ *
+ * We do a trick here to override the return type of syscall to avoid overflows.
+ */
 static inline uint64_t
 get_thread_usage_ns(void)
 {
-  uint64_t mach_time = syscall(SYS_thread_selfusage);
+  static uint64_t (*syscall64)(int, ...) = NULL;
+  if (!syscall64)
+    syscall64 = dlfcn(RTLD_NEXT, "syscall");
+
+  uint64_t mach_time = syscall64(SYS_thread_selfusage);
 
   return mach2nanos(mach_time);
 }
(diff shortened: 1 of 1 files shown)


mascguy:
@fhgwright Fred, your thoughts?

fhgwright:
@mascguy 
Sounds plausible on the surface.  I'll look into it more closely.  This is probably worth postponing the release for, since I don't think anyone is clamoring for any other 1.5.1 features.

src/time.c
```diff
@@ -685,11 +685,19 @@ get_thread_usage_ts(struct timespec *ts)
...
+ *
+ * We do a trick here to override the return type of syscall to avoid overflows.
+ */
 static inline uint64_t
 get_thread_usage_ns(void)
 {
-  uint64_t mach_time = syscall(SYS_thread_selfusage);
+  static uint64_t (*syscall64)(int, ...) = NULL;
+  if (!syscall64)
+    syscall64 = dlfcn(RTLD_NEXT, "syscall");
```
jmroot: Should this be `dlsym`? If not, where is `dlfcn` defined? If so, `dlfcn.h` probably needs to be included.

fhgwright: Yes, the function is called `dlsym`, which is declared in the `dlfcn.h` header, though I'm not yet convinced that it's necessary to use it at all in this context, and even if it were, there are macros here for that.

As I said, I'll look into it.

Un1q32:
Not sure why I write dlfcn there my bad

fhgwright:
There are multiple things wrong with this, besides the typo:

1. There's no need to use `dlsym()` at all, just to create a different prototype for an existing function.  That can be handled with the `__asm()` mechanism.
2. If `dlsym()` _were_ appropriate, it should use `GET_OS_FUNC()` or `GET_OS_ALT_FUNC()` (from `util.h`), rather than using `dlsym()` directly.
3. The approach of treating the return value as 64 bits in spite of the normal prototype doesn't work in 32-bit builds.  I haven't yet determined whether the 64-bit value is accessible in some way in 32-bit builds, or whether it will need to limit this improvement to 64-bit builds only.

I already have a test that demonstrates the failure, and a simpler fix for the 64-bit case (which the test shows doesn't work in the 32-bit case).  I'll finish the fix once I determine how to handle the 32-bit case.  Meanwhile, I suggest closing this PR.

Un1q32:
> There's no need to use dlsym() at all, just to create a different prototype for an existing function. That can be handled with the __asm() mechanism.

Never considered this.

> The approach of treating the return value as 64 bits in spite of the normal prototype doesn't work in 32-bit builds. I haven't yet determined whether the 64-bit value is accessible in some way in 32-bit builds, or whether it will need to limit this improvement to 64-bit builds only.

I'll need to reinstall macOS on my iMac to test i386, but I tested with armv7 on an iPhone 5 and it works.  On 32-bit arm, 64 bit values are passed with 2 registers, and the syscall does actually populate both registers correctly.

fhgwright:
There are comments in Apple's code that suggest that the type of syscall definition used for this function is capable of returning a 64-bit value on i386, but empirically, it doesn't work properly in this case, at least not by the simple method of pointing the 64-bit prototype at the existing function.

It doesn't even need two seconds of runtime to fail. :-)

Un1q32:
I'm reinstalling macOS on my iMac right now, once that's done I'll see what I can do.  If I can't find a solution I'll just make it use the pre-10.10 code on i386, since it still fixes the issue on x86_64.

fhgwright:
That's my fallback approach.  I'm already working on it, and as I said, I've already updated the test to be able to demonstrate the original problem.  And I have all relevant systems to test on.

The code fragment I have here is:
```
/*
 * The prototype for syscall() defines its return type as 'int', which is
 * inadequate for runtime in nanoseconds.  But the function actually
 * returns a 64-bit value in this case (on x86_64), so we can work around it by
 * supplying our own prototype, using an alternate name but referencing
 * the normal function.
 */
uint64_t __mpls_syscall64(int, ...) __asm("_syscall");
```

The function needs to be global to make the compiler happy, but that means it needs an appropriate prefix to avoid namespace pollution issues.

There's a comment in Apple's `SYS.h` that says:
```
/*
 * We have two entry points. int's is used for syscalls which need to preserve
 * %ecx across the call, or return a 64-bit value in %eax:%edx. sysenter is used
 * for the majority of syscalls which just return a value in %eax.
 */
```
This function is defined as an "int" case on i386, but %eax:%edx doesn't have an appropriate value upon return, both being significantly nonzero after a tiny bit of runtime.

Un1q32:
I found a solution, I don't know why it works but it works somehow.  Reimplementing the syscall function in assembly as
```asm
_syscall64:
  mov $0, %eax
  int $0x80
  ret
```
and calling that with a function signature of `uint64_t syscall64(int, ...)` works.

Un1q32:
```asm
_thread_selfusage:
  mov $SYS_thread_selfusage, %eax
  int $0x80
  ret
```
also works if you call with no args

Un1q32:
Here's why, this is Apple's syscall function:
```asm
_syscall:
  popl %ecx
  popl %eax
  pushl %ecx
  int $0x80
  movl (%esp), %edx
  pushl %ecx
  jae 0x19cd1
  calll 0x19cc8
  popl %edx
  movl 0x635c(%edx), %edx
  jmpl *%edx
  retl
```

64 bit values on i386 are returned in eax:edx, but right after the interrupt edx is corrupted by a mov.  By re-implementing the syscall function in a way that does not corrupt edx we can access the other bits.

Un1q32:
I'll close since I don't know how you would want me to add the assembly function into the project, so I'll just let you handle it.

fhgwright:
OK.  There's a way of writing embedded asm in C that can reference the C variable names, but it's been years since I've used it, so I'll have to look that stuff up.

It is a little bizarre that Apple clobbers that register knowing that it's part of 64-bit return values when they're used.  But it wouldn't be the first time that Apple screwed up something in 32-bit builds. :-)

Un1q32:
> It is a little bizarre that Apple clobbers that register knowing that it's part of 64-bit return values when they're used. But it wouldn't be the first time that Apple screwed up something in 32-bit builds. :-)

They probably just didn't expect/want anyone to be doing function signature overrides like this, the header declares a return value of int so trying to get a 64 bit value from it is undefined behavior.

Un1q32:
> The function needs to be global to make the compiler happy, but that means it needs an appropriate prefix to avoid namespace pollution issues.

`extern uint64_t syscall64(int, ...) __asm("_syscall");` does not create a symbol on x86 when built into a shared library in my testing, so shouldn't have any namespace issues as long as you don't name any other functions syscall64 in this project. it doesn't really matter much anyway though.

fhgwright:
@Un1q32 
It turns out that the right fix was much simpler.  I'd missed the fact that there was already a proper function wrapper for this syscall, albeit one that's undocumented and with no public prototype.  Just adding the prototype and calling it is all that's needed.  See the last commit in PR #125.

PR #128, opened 2026-02-19
Add aligned_alloc() function

programmingkidx:
Add an implementation of the aligned_alloc() function.

### DIFF
--- a/include/stdlib.h
+++ b/include/stdlib.h
@@ -86,6 +86,8 @@ __MP__BEGIN_DECLS
 extern void arc4random_buf( void* buf, size_t n );
 __MP__END_DECLS
 
+void *aligned_alloc(size_t requested_alignment, size_t size);
+
 #endif /*  __MPLS_SDK_SUPPORT_ARC4RANDOM__ */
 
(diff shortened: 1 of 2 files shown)


mascguy:
Adding @fhgwright for review/comment.

fhgwright:
@programmingkidx 
This has many issues:

1. It doesn't follow the normal structure for `legacy-support` features.
2. It doesn't include any kind of test.
3. It includes some debugging code that's redundant given that it (inappropriately) wraps `malloc()`.
4. Most importantly, because it just uses `malloc()`, it doesn't actually guarantee any alignment beyond the usual 16 bytes, which is supposed to be the whole point of the function.

It's also worth noting that Apple limits its availabllity to C/C++ versions that have it as part of the language standard, presumably to avoid namespace pollution.

I added an implementation in #132 which addresses all of the above, but includes a flag to force its availability in earlier C/C++ versions.  It's not clear how much value this has, given that it's just a somewhat more restrictive version of `posix_memalign()`.

See: [https://github.com/macports/macports-legacy-support/pull/132/commits/70c72687d67466c655f13dd21d8691f4445a1e60](url).

PR #131, opened 2026-07-14
Add CCRandomGenerateBytes(), in CommonCrypto/CommonRandom.h

nilFinx:
Untested if CommonCrypto will continue to work fine on 10.13+, in case LegacySystem is being used. It would help if someone else could...

Also, excuse me for any consistency, or other code issues. This is my first time contributing any form of C code.


mascguy:
Adding @fhgwright for review/comment.

fhgwright:
@nilFinx 
I see that you did take the time to study some existing code and get a sense of the conventions, though that was a bit overzealous in thinking that `add_symbols.c` should be updated. :-)

That particular file is an ugly kludge for a narrow purpose, and shouldn't be updated.

General structural comments:

All nontrivial files should have copyright notices (though they don't necessarily need a personal name).  It's incorrect to assume that open-source code doesn't need copyrights, because without a copyright, some other party could come along and add their own and demand license fees.  So the "free" part needs to be stated explicitly.  Even Richard Stallman understands that.

A comment to say why something is uncommented with almost as much text as the omitted comment is pretty silly. :-)

There's no test included.  All new features should include tests for at least a basic verification that they're usable.  There are some existing features without tests, but their number should hopefully decrease with time, and certainly shouldn't increase.  There are some unwritten guidelines for tests, though it's not worth spelling them out here (see below).

Functional comments:

The `CCRandomGenerateBytes` function was added in 10.10, so 10.13 is the wrong threshold.

The `CommonRandom.h` header was not present at all prior to 10.10, so the `include_next` will fail in all <10.10 cases.  In this particular case, the threshold for the appearance of the header matches the appearance of the feature, so they can just use complementary states of the same flag, though that isn't necessarily true in general.

The apparently infinitely recursive macro definition isn't the proper way to provide a function prototype.  If `cpp` didn't disable recursion, any use of this macro would crash `cpp` with a stack overflow.  And in general, using macros for prototypes is potentially problematic and should be avoided as much as possible.  The usual way to provide the prototype for a `legacy-support` function is to copy Apple's and delete the AVAILABILITY stuff.

It's missing the `__MP_xxx_DECLS` wrapper, which is needed to make the function callable from C++.

And now the biggie:

The `CCRandomGenerateBytes` function is expected to provide cryptographic-quality randomness, complete with FIPS compliance.  This is not true of `arc4random`.  Substituting a lower-grade randomness function for one expected to provide cryptographic-quality randomness is dangerous and could lead to security problems.

In principle, backporting Apple's implementation to older OS versions would be a good approach.  It's just one code source file, but it uses a number of headers, some of which come from the kernel and hence suggest kernel involvement.  And the kernel-provided headers don't even exist prior to 10.8, complicating full-range support.  I wouldn't rule out this approach, but there would be significant effort involved in understanding and sorting out the issues.

The bottom line is that a _proper_ implementation of `CCRandomGenerateBytes` is substantially more involved than just providing a wrapper around `arc4random`.  Though any program satisfied with the quality of `arc4random` is free to use it on its own.

nilFinx:
Oh - I was originally plamming to make this a fork, and not actually merged. Sorry for the "freeware lmao" ccomment caused by that.

I'll force push and improve it, but I understand that arc4random is not that secure/random.

I'll see what I can do from now.

fhgwright:
@nilFinx 
Obviously you can do whatever you want in your own fork, but a PR is an indication of a desire to merge it into the target repo.  It can be marked as a draft to indicate that it's not _yet_ ready to be merged, but even draft PRs are assumed to be intended for _eventual_ merging.

I don't think it's appropriate to merge anything based on `arc4random`, for the reason stated above.

fhgwright:
@nilFinx 
I decided that it's OK to offer the less secure version as long as it's explicitly requested with a flag.  The implementation uses a different name so that it can't be used accidentally just by not bothering with the prototype, and the optional prototype maps the name.

See: [https://github.com/macports/macports-legacy-support/pull/132/commits/90d51e7f0baed8f0f614ff78bc190682a65f387f](url).

nilFinx:
I want to look into this again, as quite a number of Rust stuff depends on it.
It appears that OpenSSL might be secure enough if there are checks to ensure that the operation is safe enough. How about using that?

fhgwright:
@nilFinx 
> I want to look into this again, as quite a number of Rust stuff depends on it. It appears that OpenSSL might be secure enough if there are checks to ensure that the operation is safe enough. How about using that?

Anything that wants to substitute OpenSSL code is free to do so, and it's probably not very hard.  But there's no way in hell that it would be reasonable to make OpenSSL a dependency of legacy-support.

